PR #18, opened 2022-06-23
prepare 1.0.1 release

LaunchDarklyReleaseBot:
### Changed:
- Update the example project to a new version of flutter embedding and removed usage of deprecated flutter components.
- Update to use ios-client-sdk V6.
- Updated to work with kotlin 1.7.0.

### Fixed:
- `identify` calls were blocking and could trigger ANRs. They maintain the same interface, and can be awaited, but now they no longer block the calling thread.

### DIFF
--- a/.ldrelease/config.yml
+++ b/.ldrelease/config.yml
@@ -13,9 +13,9 @@ jobs:
       context: org-global
       timeoutMinutes: 20
       mac:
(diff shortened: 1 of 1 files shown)


louis-launchdarkly (approved):

kinyoklion:
Changing version.

PR #61, opened 2024-01-19
feat: contract test coverage for private attributes, context build an…

tanderson-ld:
Added more contract test coverage.

### DIFF
--- a/apps/flutter_client_contract_test_service/README.md
+++ b/apps/flutter_client_contract_test_service/README.md
@@ -20,12 +20,4 @@ If you need to add or modify the OpenAPI yaml files to add or modify functionali
 dart run build_runner build -v
 ```
 
-You will also need to patch service_api.openapi.dart to deal with a `Object` vs `dynamic` typing bug.
-
-```
-replace...
-final Map<String, Object> _additionalProperties = <String, Object>{};
-
-with...
-final Map<String, dynamic> _additionalProperties = <String, dynamic>{};
-```
+You will also need to patch service_api.openapi.dart by replacing all instances of `Object` vs `dynamic`.  This is due to a bug in the generator implementation.
--- a/apps/flutter_client_contract_test_service/pubspec.lock
+++ b/apps/flutter_client_contract_test_service/pubspec.lock
@@ -435,18 +435,18 @@ packages:
     dependency: transitive
     description:
       name: material_color_utilities
-      sha256: "0e0a020085b65b6083975e499759762399b4475f766c21668c4ecca34ea74e5a"
+      sha256: "9528f2f296073ff54cb9fee677df673ace1218163c3bc7628093e7eed5203d41"
       url: "https://pub.dev"
     source: hosted
-    version: "0.8.0"
+    version: "0.5.0"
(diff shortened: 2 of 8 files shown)


apps/flutter_client_contract_test_service/README.md
```diff
@@ -20,12 +20,4 @@ If you need to add or modify the OpenAPI yaml files to add or modify functionali
...
-You will also need to patch service_api.openapi.dart to deal with a `Object` vs `dynamic` typing bug.
-
-```
-replace...
-final Map<String, Object> _additionalProperties = <String, Object>{};
-
-with...
-final Map<String, dynamic> _additionalProperties = <String, dynamic>{};
-```
+You will also need to patch service_api.openapi.dart by replacing all instances of `Object` vs `dynamic`.  This is due to a bug in the generator implementation.
```
kinyoklion, line 23: Is there an issue for this in the generator repo? We should probably file one if not. Or maybe we will get an alternative.

apps/flutter_client_contract_test_service/pubspec.lock
```diff
@@ -435,18 +435,18 @@ packages:
     dependency: transitive
     description:
       name: material_color_utilities
-      sha256: "0e0a020085b65b6083975e499759762399b4475f766c21668c4ecca34ea74e5a"
+      sha256: "9528f2f296073ff54cb9fee677df673ace1218163c3bc7628093e7eed5203d41"
       url: "https://pub.dev"
     source: hosted
-    version: "0.8.0"
+    version: "0.5.0"
```
kinyoklion, line 441: Strange to see them go down.

apps/flutter_client_contract_test_service/testharness-suppressions.txt
```diff
@@ -1,22 +1,11 @@
-events/context properties/single-kind minimal
-events/context properties/multi-kind minimal
-events/context properties/single-kind with attributes, nothing private
-events/context properties/single-kind, allAttributesPrivate
-events/context properties/single-kind, specific private attributes
-events/context properties/single-kind, private attribute nested property
 events/disabling/evaluation
```
kinyoklion, line 1: Does this pass after the dynamic change. It seems like it should.

kinyoklion (approved):

apps/flutter_client_contract_test_service/testharness-suppressions.txt
kinyoklion: I gave it a try and it did.

tanderson-ld: Removed from suppression.  Thanks.

apps/flutter_client_contract_test_service/pubspec.lock
tanderson-ld: Reverted.  I'm unsure why this happens on my machine.

apps/flutter_client_contract_test_service/README.md
kinyoklion: :+1:

PR #64, opened 2024-01-22
chore: Preparation for alpha packages.

kinyoklion:
This meets some publishing expectations and standards.

Each package needs a LICENSE.

The entrypoint in the package should match the name of the package.


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2023 Catamorphic, Co.
```
kinyoklion, line 1: Normally I don't change these dates, but we are publishing it now, so it seemed reasonable.

packages/flutter_client_sdk/README.md
```diff
@@ -11,6 +11,12 @@
 
 [![Twitter Follow](https://img.shields.io/twitter/follow/launchdarkly.svg?style=social&label=Follow&maxAge=2592000)](https://twitter.com/intent/follow?screen_name=launchdarkly)
 
+## Platform Support
```
kinyoklion, line 14: This seems to be the way most packages represent this, so I added it here. The caveats about compatibility are in the top level README.md. Should be sufficient.

tanderson-ld (approved):

PR #65, opened 2024-01-22
fix: Remove flutter dependency from event source.

kinyoklion:


packages/flutter_client_sdk/lib/src/ld_client.dart
```diff
@@ -68,7 +68,7 @@ interface class LDClient {
         context,
         DiagnosticSdkData(
             name: 'FlutterClientSide',
-            version: '0.0.1')); // x-release-please-version
+            version: '4.0.0-alpha.0')); // x-release-please-version
```
kinyoklion, line 71: Went ahead and updated this.

packages/flutter_client_sdk/pubspec.yaml
kinyoklion, line 1: I've not published this one yet, so it can stay at `.0`.

tanderson-ld (approved):

PR #66, opened 2024-01-22
fix: Fix sink not closed lint.

kinyoklion:
Updated lints to 3.0+.
Fixed lint pointed out by publish.

Updated lint rules to detect this locally.

Suppressed issue in tests.


kinyoklion:
Moved back to draft to add more analysis rules.

tanderson-ld (approved):

PR #69, opened 2024-01-23
build(deps): Allow uuid 3.x and 4.x

kinyoklion:
The API used by the SDK should be the same with 3.x and 4.x of UUID.


kinyoklion:
The CI will fail on format until the other PR is merged.

kinyoklion:
I tested out an example app with the 4.x dependency.

tanderson-ld (approved):

PR #95, opened 2024-01-25
fix: Check for network state on resume.

kinyoklion:
Description in the comments inside the changes.

### DIFF
--- a/packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
+++ b/packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
@@ -74,6 +74,21 @@ final class FlutterStateDetector implements StateDetector {
       case AppLifecycleState.detached:
         break;
       case AppLifecycleState.resumed:
+        // From connectivity_plus:
+        //
+        // "Connectivity changes are no longer communicated to Android apps in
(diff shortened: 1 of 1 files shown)


packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
```diff
@@ -74,6 +74,21 @@ final class FlutterStateDetector implements StateDetector {
...
+        // only useful when your application is in the foreground."
+        // https://github.com/fluttercommunity/plus_plugins/tree/main/packages/connectivity_plus/connectivity_plus
+        //
+        // So, when we detect an that we have been resumed, we query the current
+        // connectivity state and emit an event.
+        // There is some excess checking here, especially during app load, but
+        // our reaction to the state depends on it being different, so reporting
+        // the same state, in excess, has minimal impact and is better than
+        // missing an active state transition.
+        Connectivity().checkConnectivity().then(_setConnectivity);
```
kinyoklion, line 91: There is, at an interface level, a race condition here. Which is if this completes immediately after we start getting an event with conflicting information.

Now sure it is an actual race condition at an implementation level. 

Workarounds mostly have the same or different problem.

keelerm84 (approved):

PR #96, opened 2024-01-25
feat: Implement support for REPORT for streaming.

kinyoklion:
Implement REPORT support in the event source.
Update the SSE contract tests with more capabilities.
Update the streaming data source to support streaming report.

Some misc removals and cleanups of minor todo items.

### DIFF
--- a/apps/flutter_client_contract_test_service/testharness-suppressions.txt
+++ b/apps/flutter_client_contract_test_service/testharness-suppressions.txt
@@ -1,10 +1,2 @@
-polling/requests/context properties/single kind with all attributes/REPORT
-polling/requests/context properties/multi-kind/REPORT
-streaming/requests/method and headers/REPORT
-streaming/requests/URL path is computed correctly/no environment filter/base URI has no trailing slash/REPORT
-streaming/requests/URL path is computed correctly/no environment filter/base URI has a trailing slash/REPORT
-streaming/requests/context properties/single kind minimal/REPORT
-streaming/requests/context properties/single kind with all attributes/REPORT
-streaming/requests/context properties/multi-kind/REPORT
 context type/convert/invalid context/{"kind": null, "key": "x"}
 context type/compare
--- a/apps/sse_contract_test_service/bin/service_api.openapi.dart
+++ b/apps/sse_contract_test_service/bin/service_api.openapi.dart
@@ -5,6 +5,7 @@
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openapi_base/openapi_base.dart';
+
 part 'service_api.openapi.g.dart';
 
 @JsonSerializable()
@@ -22,6 +23,7 @@ class ServiceStatusResponse implements OpenApiContent {
   final List<String>? capabilities;
 
   Map<String, dynamic> toJson() => _$ServiceStatusResponseToJson(this);
+
   @override
   String toString() => toJson().toString();
 }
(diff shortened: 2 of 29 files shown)


apps/flutter_client_contract_test_service/testharness-suppressions.txt
```diff
@@ -1,10 +1,2 @@
-polling/requests/context properties/single kind with all attributes/REPORT
```
kinyoklion, line 1: These first two were already working.

apps/sse_contract_test_service/bin/service_api.openapi.yaml
```diff
@@ -75,8 +75,7 @@ components:
           type: string
         headers:
           type: object
-          additionalProperties:
```
kinyoklion, line 78: So, the generator works completely wrong in this case. It works 50% wrong with `true`. So the generated code has been manually edited.

apps/sse_contract_test_service/bin/sse_contract_test_service.dart
```diff
@@ -14,7 +14,9 @@ class TestApiImpl extends TestApi {
 
   @override
   Future<GetResponse> Get() async {
-    return GetResponse.response200(ServiceStatusResponse());
+    const capabilities = <String>['report', 'post', 'headers'];
```
kinyoklion, line 17: It may support some more capabilities, but I am sure it supports these.

apps/sse_contract_test_service/bin/service_api.openapi.dart
```diff
@@ -31,12 +33,30 @@ class ServiceStatusResponse implements OpenApiContent {
 class CreateStreamRequestHeaders implements OpenApiContent {
   CreateStreamRequestHeaders();
 
+  // This type has been manually edited to be a correct string to string map.
+  // The generator either omitted all properties with a correctly formed
+  // schema, or generated incorrect code with an untyped map.
   factory CreateStreamRequestHeaders.fromJson(Map<String, dynamic> jsonMap) =>
-      _$CreateStreamRequestHeadersFromJson(jsonMap);
+      _$CreateStreamRequestHeadersFromJson(jsonMap)
```
kinyoklion, line 40: In this factory the code was manually changed from `Object` to `String`.

packages/event_source_client/lib/launchdarkly_event_source_client.dart
```diff
@@ -11,10 +11,21 @@ import 'src/sse_client_stub.dart'
 
 export 'src/message_event.dart' show MessageEvent;
 
-// TODO: sc-157779 - add general logging
```
kinyoklion, line 14: One TODO fixed, the other TODOs moved to shortcut.

packages/common/lib/src/serialization/ld_value_serialization.dart
```diff
@@ -30,9 +30,6 @@ final class LDValueSerialization {
       return LDValue.ofNull();
     }
 
-    // TODO: We may want to consider an error. If we do, then we would basically
```
kinyoklion, line 33: I don't think there is a great deal of value here, so I am removing this.

packages/common_client/lib/src/data_sources/data_source_event_handler.dart
```diff
@@ -104,7 +104,6 @@ final class DataSourceEventHandler {
       _statusManager.setValid();
       return MessageStatus.messageHandled;
     } catch (err) {
-      // TODO sc-215195: fix contract test that is failing here due to empty flag content
```
kinyoklion, line 107: This was not why contract tests failed here. It was because the data was invalid. For a client SDK an empty data set is `{}` for a server is it `{"flags":{},"segments":{}}`, which is an invalid payload. This has been fixed in the contract tests.

packages/flutter_client_sdk/example/pubspec.yaml
```diff
@@ -30,13 +30,10 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  # This defaults to a published package. To use the local SDK comment the following line out,
-  # and then uncomment the two lines after. This default is to make it easier to run the example
-  # without needing to build the SDK locally.
+  # This defaults to a published package. Using `flutter pub get` will get the package from pub.dev
+  #If you run `melos bs` in the root of the repository, then it will be linked to the local version
```
keelerm84, line 34:
```suggestion
  # If you run `melos bs` in the root of the repository, then it will be linked to the local version
```

keelerm84 (approved):

PR #98, opened 2024-01-25
Rlamb/single environment collection

kinyoklion:
Collect environment information once, and calculate recipes once.

I think the recipe could be made simpler, but that is outside the immediate scope.

### DIFF
--- a/packages/common/lib/src/env_reporting/environment_reporter.dart
+++ b/packages/common/lib/src/env_reporting/environment_reporter.dart
@@ -2,6 +2,16 @@ import '../application_info.dart';
 import '../os_info.dart';
 import '../device_info.dart';
 
+final class EnvironmentReport {
+  ApplicationInfo? applicationInfo;
+  OsInfo? osInfo;
+  DeviceInfo? deviceInfo;
+  String? locale;
+
+  EnvironmentReport(
+      {this.applicationInfo, this.osInfo, this.deviceInfo, this.locale});
+}
+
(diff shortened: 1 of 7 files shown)


packages/common/lib/src/env_reporting/environment_reporter.dart
```diff
@@ -2,6 +2,16 @@ import '../application_info.dart';
 import '../os_info.dart';
 import '../device_info.dart';
 
+final class EnvironmentReport {
```
kinyoklion, line 5: A constructed report which doesn't require further async work.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -221,19 +221,19 @@ final class LDCommonClient {
     // TODO: Do we start the process when we create the client, and provide
     // a way to know when it completes? Or do we not even start it as we
     // are doing here.
-    _envReporter = await _makeEnvReporter();
+    _envReport = await _makeEnvironmentReport();
```
tanderson-ld, line 224: Is this running in all cases now?  Before I believe if a customer had auto env attributes turned off, the env code would only query for the osInfo to use in the diagnostic portion down below.

kinyoklion: It does. We should make sure that we are clear that this information only is contingent on going into the context, not that we don't call the APIs.

tanderson-ld (approved):

PR #102, opened 2024-01-26
feat: Add restart support to the SSE client.

kinyoklion:
This adds restart support to the SSE client. We can then use this to restart the client on bad data and still include backoff/jitter.

### DIFF
--- a/apps/sse_contract_test_service/bin/sse_contract_test_service.dart
+++ b/apps/sse_contract_test_service/bin/sse_contract_test_service.dart
@@ -7,14 +7,15 @@ import 'service_api.openapi.dart';
 import 'dart:io';
 
 class TestApiImpl extends TestApi {
-  static const clientUrlPrefix = "/client/";
+  static const _clientUrlPrefix = "/client/";
 
-  final Map<int, StreamSubscription> clientSubMap = {};
-  var nextIdToGive = 0;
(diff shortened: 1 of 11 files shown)


apps/sse_contract_test_service/bin/sse_contract_test_service.dart
```diff
@@ -7,14 +7,15 @@ import 'service_api.openapi.dart';
 import 'dart:io';
 
 class TestApiImpl extends TestApi {
-  static const clientUrlPrefix = "/client/";
+  static const _clientUrlPrefix = "/client/";
```
kinyoklion, line 10: Changed these all to private variables. Unrelated to changes.

packages/event_source_client/lib/src/sse_client_html.dart
```diff
@@ -49,6 +49,11 @@ class HtmlSseClient implements SSEClient {
 
   @override
   Future close() => _messageEventsController.close();
+
+  @override
+  void restart() {
+    // TODO: implement restart
```
kinyoklion, line 55: Not implementing this right now. We need to implement jitter/backoff for the web first.

louis-launchdarkly (approved):

tanderson-ld (approved):

PR #104, opened 2024-01-29
fix: anonymous context keys and auto env context keys are now in separate namespaces

tanderson-ld:
**Related issues**

https://app.shortcut.com/launchdarkly/story/231552/ensure-device-keys-are-a-different-namespace-than-anonymous-keys


kinyoklion (approved):

kinyoklion (changes requested):
Need to dart format.

kinyoklion (changes requested):
Looks like the unit tests need updated as well.

kinyoklion (approved):

PR #106, opened 2024-01-29
fix: making context builder easier to use

tanderson-ld:
Adds primitive type setters to the context builder API to make it more user friendly.

**Related issues**

https://app.shortcut.com/launchdarkly/story/222279/make-the-api-for-contexts-and-values-more-uniform-ideally-with-specific-typed-methods-in-addition-to-an-ldvalue-accepting

### DIFF
--- a/packages/common/lib/src/ld_context.dart
+++ b/packages/common/lib/src/ld_context.dart
@@ -101,6 +101,12 @@ final class LDAttributesBuilder {
   LDAttributesBuilder kind(String kind, [String? key]) =>
       _parent.kind(kind, key);
 
+  /// Set the name of the context.
+  LDAttributesBuilder name(String name) {
+    _name = name;
+    return this;
+  }
+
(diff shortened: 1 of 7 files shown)


packages/common/lib/src/ld_context.dart
```diff
@@ -143,20 +150,30 @@ final class LDAttributesBuilder {
   /// in the LaunchDarkly model; any expressions in feature flags that reference
   /// an attribute with a null value will behave as if the attribute did not
   /// exist.
-  LDAttributesBuilder set(String name, LDValue value) {
-    _trySet(name, value);
-
+  LDAttributesBuilder setValue(String name, LDValue value,
```
kinyoklion, line 153: I have some hesitation in changing this, but I guess the upgrade will already be extensive.

kinyoklion:
This one will be important to @mmrj I think.

kinyoklion (approved):

PR #107, opened 2024-01-29
fix: Mac entitlements. Clarifications.

kinyoklion:


packages/flutter_client_sdk/example/macos/Runner/DebugProfile.entitlements
```diff
@@ -8,5 +8,7 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+	<key>com.apple.security.network.client</key>
```
kinyoklion, line 11: Needed to make network requests.

packages/flutter_client_sdk/example/README.md
```diff
@@ -18,3 +18,6 @@ mobile key or client-side ID.
 
 When running from an IDE you can edit the build configuration to include the `--dart-define` commands.
 In Android studio these can be placed in the "Additional Run Args" when you edit the run configuration.
+
+In order to run the example on a physical iOS device you will need to configure a development team and provisioning profile.
```
kinyoklion, line 22: Not going to include details, as those change all the time, and an iOS developer should be somewhat familiar with the process.

PR #113, opened 2024-01-30
fix: Flutter client should depend only on common client.

kinyoklion:
I noticed the flutter SDK directly depended on common. This was because common client was missing some re-exports. This fixes that. So I will release the patch of common client after merging this. Then do the final version number updates.

### DIFF
--- a/packages/flutter_client_sdk/pubspec.yaml
+++ b/packages/flutter_client_sdk/pubspec.yaml
@@ -14,8 +14,7 @@ dependencies:
     sdk: flutter
   package_info_plus: ^5.0.1
   device_info_plus: ^9.1.1
-  launchdarkly_common_client: 0.1.0
-  launchdarkly_dart_common: 0.0.2
+  launchdarkly_common_client: 1.0.0
   shared_preferences: ^2.2.2
   connectivity_plus: ^5.0.2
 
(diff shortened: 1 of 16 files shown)


packages/flutter_client_sdk/pubspec.yaml
```diff
@@ -14,8 +14,7 @@ dependencies:
     sdk: flutter
   package_info_plus: ^5.0.1
   device_info_plus: ^9.1.1
-  launchdarkly_common_client: 0.1.0
-  launchdarkly_dart_common: 0.0.2
```
kinyoklion, line 18: Removing this.

packages/flutter_client_sdk/pubspec.yaml
```diff
@@ -14,8 +14,7 @@ dependencies:
...
+  launchdarkly_common_client: 1.0.0
```
kinyoklion, line 17: This will need updated again after the publish.

release-please-config.json
```diff
@@ -3,16 +3,13 @@
   "separate-pull-requests": true,
   "packages": {
     "packages/common": {
```
kinyoklion, line 5: All, the 1.0.0 have been released, so now those should auto-bump from now on.

louis-launchdarkly (approved):

PR #135, opened 2024-03-01
feat: Add LDContext equality comparison.

kinyoklion:


packages/common/lib/src/ld_context.dart
```diff
@@ -75,6 +76,27 @@ final class LDContextAttributes {
...
+          customAttributes.equals(other.customAttributes) &&
+          kind == other.kind &&
+          key == other.key &&
+          anonymous == other.anonymous &&
+          name == other.name &&
+          privateAttributes.equals(other.privateAttributes);
+
+  @override
+  int get hashCode =>
+      Object.hashAllUnordered(customAttributes.keys) ^
```
tanderson-ld, line 93: Well that's a very useful function.

tanderson-ld (approved):

packages/common/lib/src/ld_context.dart
kinyoklion: It is indeed fantastic.

PR #137, opened 2024-03-05
chore: Support context comparison contract test.

kinyoklion:


apps/flutter_client_contract_test_service/bin/contract_test_service.dart
```diff
@@ -298,6 +297,61 @@ class TestApiImpl extends SdkTestApi {
     }
   }
 
+  LDContext _buildContextForComparison(SingleOrMultiBuildContext input) {
```
kinyoklion, line 300: Contexts for comparison are not built the same was as contexts for conversions or context builds. We may need to adjust them as well, but they have different considerations so the various types are different.

tanderson-ld (approved):

PR #148, opened 2024-03-26
feat: Support latest versions of the plus plugins.

maxlapides:
**Requirements**

- [x] I have validated my changes against all supported platform versions

**Related issues**

It's not currently possible to upgrade the Flutter plus plugins in projects that require the LaunchDarkly SDK until the LaunchDarkly SDK upgrades its dependencies.

Upgraded the dependencies.

N/A

**Additional context**

N/A

### DIFF
--- a/packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
+++ b/packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
@@ -23,7 +23,8 @@ final class FlutterStateDetector implements StateDetector {
   Stream<NetworkState> get networkState => _networkStateController.stream;
 
   late final LDAppLifecycleListener _lifecycleListener;
-  late final StreamSubscription<ConnectivityResult> _connectivitySubscription;
+  late final StreamSubscription<List<ConnectivityResult>>
+      _connectivitySubscription;
 
   FlutterStateDetector() {
     final initialState = SchedulerBinding.instance.lifecycleState;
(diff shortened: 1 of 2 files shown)


packages/flutter_client_sdk/pubspec.yaml
```diff
@@ -12,11 +12,11 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  package_info_plus: ">=4.2.0 <6.0.0"
-  device_info_plus: ^9.1.1
```
kinyoklion, line 16: It looks like only connectivity plus introduced changes? So this one could be ">=9.1.1 <11.0.0"?

packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
```diff
@@ -23,7 +23,8 @@ final class FlutterStateDetector implements StateDetector {
   Stream<NetworkState> get networkState => _networkStateController.stream;
 
   late final LDAppLifecycleListener _lifecycleListener;
-  late final StreamSubscription<ConnectivityResult> _connectivitySubscription;
```
kinyoklion, line 26: I am a little concerned about also requiring an upgrade to a package that has only been out a few days. 

Ideally we could support both for a bit. Probably not elegant, but I wonder if we could wrap these couple contact points. Accept a dynamic, if the item is not a list, then turn it into a list of one, if it is a list then forward it along.

kinyoklion:
@maxlapides Thank you for the contribution! I've added a couple comments, and we will need a little time to consider.

kinyoklion:
I changed the PR title to ensure that it is releasable via release-please. It is a "feat" which would correspond to a minor version bump.

packages/flutter_client_sdk/pubspec.yaml
maxlapides: Yes, good idea!

packages/flutter_client_sdk/lib/src/flutter_state_detector.dart
maxlapides: Definitely not elegant, but it's not a bad idea! I just updated this PR, it should do the trick.

I thought about writing a wrapper as you suggested, but it looks like it's really just this one little spot in the code that needs to handle connectivity_plus v5, so I thought I'd keep it simple.

kinyoklion: @maxlapides Looks great. Thank you for the conrtibution!

kinyoklion (approved):

maxlapides:
Thanks so much @kinyoklion!

PR #154, opened 2024-04-15
feat: Support package_info_plus 7.x

kinyoklion:


packages/flutter_client_sdk/pubspec.yaml
```diff
@@ -12,7 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  package_info_plus: ">=4.2.0 <7.0.0"
```
kinyoklion, line 15: The changes for 7.0.0 don't affect the API, but they did depend on a new API. So if someone wanted to use 7.0.0 they would need to use at least flutter 3.19.

tanderson-ld (approved):

PR #157, opened 2024-05-02
feat: Support package_info_plus 8.

kinyoklion:


packages/flutter_client_sdk/example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -11,7 +11,7 @@ import package_info_plus
 import shared_preferences_foundation
 
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
-  ConnectivityPlugin.register(with: registry.registrar(forPlugin: "ConnectivityPlugin"))
```
kinyoklion, line 14: I think this is innocuous and maybe associated with running a new patch version.

tanderson-ld (approved):

PR #159, opened 2024-05-09
chore: Add the startWithTimeout method to ensure a timeout is used

louis-launchdarkly:
Because the current start method returns a future, to not create behavioral change, it probably makes sense to create a new method to use.


tanderson-ld (changes requested):

kinyoklion (changes requested):
I am confused here. I thought we were an optional timeout to the one method, it warns you when you don't use it. Next major version we make the timeout either mandatory or default to a value when not specified.

Requesting changes as to not lose track.

PR #160, opened 2024-07-24
feat: Add support for waiting for non-cached values.

kinyoklion:
1. Allows for the common client to have customized data sources. This was used for testing here, but could be used for more specialized use-cases in the future.
2. Adds the ability for `start` and `identify` to wait for values from LaunchDarkly instead of resolving with cached values when available.

With the previous wrapper SDK this behavior was controlled by the underlying SDK and those would operate similar to the new, optional, behavior.

For most uses cases it is both safer and higher performance to used cached values when they are available instead of only using those cached values as a fallback.

Ideally we would add new return values to start and identify. Next major version we should consider:
`Cached`, `Timeout`. While retaining the current value for completing via data from LD.

### DIFF
--- a/packages/common_client/lib/src/ld_common_client.dart
+++ b/packages/common_client/lib/src/ld_common_client.dart
@@ -45,6 +45,37 @@ final class IdentifyError implements IdentifyResult {
   IdentifyError(this.error);
 }
 
+typedef DataSourceFactoriesFn = Map<ConnectionMode, DataSourceFactory> Function(LDCommonConfig config, LDLogger logger, HttpProperties httpProperties);
+
+Map<ConnectionMode, DataSourceFactory> _defaultFactories(LDCommonConfig config, LDLogger logger, HttpProperties httpProperties) {
+  return {
+    ConnectionMode.streaming: (LDContext context) {
+      return StreamingDataSource(
+          credential: config.sdkCredential,
+          context: context,
+          endpoints: config.serviceEndpoints,
+          logger: logger,
+          dataSourceConfig: StreamingDataSourceConfig(
+              useReport: config.dataSourceConfig.useReport,
+              withReasons: config.dataSourceConfig.evaluationReasons),
+          httpProperties: httpProperties);
+    },
+    ConnectionMode.polling: (LDContext context) {
+      return PollingDataSource(
+          credential: config.sdkCredential,
+          context: context,
+          endpoints: config.serviceEndpoints,
+          logger: logger,
+          dataSourceConfig: PollingDataSourceConfig(
+              useReport: config.dataSourceConfig.useReport,
+              withReasons: config.dataSourceConfig.evaluationReasons,
+              pollingInterval:
+              config.dataSourceConfig.polling.pollingInterval),
+          httpProperties: httpProperties);
+    },
+  };
+}
+
 final class LDCommonClient {
   final LDCommonConfig _config;
   final Persistence _persistence;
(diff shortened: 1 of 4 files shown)


packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -375,19 +396,19 @@ final class LDCommonClient {
...
     final loadedFromCache = await _flagManager.loadCached(_context);
 
     if (_config.offline) {
-      // TODO: Do we need to do anything different here?
       return;
     }
     _dataSourceManager.identify(_context, completer);
 
-    if (loadedFromCache) {
+    if (loadedFromCache && !waitForNonCachedValues) {
```
kinyoklion, line 411: The actual change to identify and start logic.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -398,7 +419,7 @@ final class LDCommonClient {
   /// Will return the provided [defaultValue] if the flag is missing, not a bool, or if some error occurs.
```
kinyoklion, line 419: Remainder are incidental formatting changes.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -45,6 +45,37 @@ final class IdentifyError implements IdentifyResult {
   IdentifyError(this.error);
 }
 
+typedef DataSourceFactoriesFn = Map<ConnectionMode, DataSourceFactory> Function(LDCommonConfig config, LDLogger logger, HttpProperties httpProperties);
+
+Map<ConnectionMode, DataSourceFactory> _defaultFactories(LDCommonConfig config, LDLogger logger, HttpProperties httpProperties) {
```
kinyoklion, line 50: Moved the default factories into a function. Now it will choose between provided and defaults.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -179,7 +215,15 @@ final class LDCommonClient {
...
-  Future<bool> start() {
+  ///
+  /// The [waitForNonCachedValues] parameters, when true, indicates that the SDK
+  /// will attempt to wait for values from LaunchDarkly instead of depending
+  /// on cached values. The cached values will still be loaded, but the future
+  /// returned by this function will not resolve. Generally this
+  /// option should NOT be used and instead flag changes should be listened to.
+  /// If [waitForNonCachedValues] is true, and an error is encountered, then
+  /// false may be returned even if cached values were loaded.
+  Future<bool> start({bool waitForNonCachedValues = false}) {
```
kinyoklion, line 226: So, if you want to use cached values on app start, but not on identify, then this supports that.

If you want to identify multiple contexts sequentially, and you want to ensure your cache gets updated, you can do that. You can also choose to use the cached values for the last context, being as that one would be updated in the background.

If you want to ensure you have "fresh" values when navigating to specific sections of your app, you could do that.

What this does not solve is the situation where someone wants to identify multiple contexts sequentially, wants the cache to be updated, but doesn't want the performance penalty. (One request was basically to allow the identify to continue and update the cache in the background, but concurrently start identifying a new context.)

kinyoklion: If you want flag values to "not change", then you can evaluate all relevant flags after the identify and use those values.

Alternatively, you could set the SDK to offline after completing the identify, and then only go online again when you want to specifically update values.

tanderson-ld (approved):

PR #166, opened 2024-10-07
fix: Use correct flag version in evaluation events.

kinyoklion:
There is an associated change to the sdk-test harness to ensure we don't encounter this again.

### DIFF
--- a/apps/flutter_client_contract_test_service/pubspec.lock
+++ b/apps/flutter_client_contract_test_service/pubspec.lock
@@ -398,7 +398,7 @@ packages:
       path: "../../packages/flutter_client_sdk"
       relative: true
     source: path
-    version: "4.6.0"
+    version: "4.7.0"
   lints:
     dependency: "direct dev"
     description:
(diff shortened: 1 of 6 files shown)


apps/flutter_client_contract_test_service/pubspec.lock
```diff
@@ -398,7 +398,7 @@ packages:
       path: "../../packages/flutter_client_sdk"
       relative: true
     source: path
-    version: "4.6.0"
+    version: "4.7.0"
```
kinyoklion, line 401: Change from `melos bs`.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -551,7 +551,7 @@ final class LDCommonClient {
             ? DateTime.fromMillisecondsSinceEpoch(
                 evalResult!.flag!.debugEventsUntilDate!)
             : null,
-        version: evalResult?.version));
+        version: evalResult?.flag?.flagVersion ?? evalResult?.version));
```
kinyoklion, line 554: Primary change of importance.

tanderson-ld (approved):

PR #172, opened 2024-10-30
feat: Adds support for client-side prerequisite events

tanderson-ld:
**Requirements**

- [ ] I have validated my changes against all supported platform versions
Exception: Changes are not related to platforms.

**Related issues**

SDK-691

### DIFF
--- a/packages/common/lib/src/events/default_event_processor.dart
+++ b/packages/common/lib/src/events/default_event_processor.dart
@@ -50,7 +50,7 @@ final class DefaultEventProcessor implements EventProcessor {
 
   DefaultEventProcessor(
       {required LDLogger logger,
-      bool indexEvents = false,
+      required bool indexEvents,
       required int eventCapacity,
       required Duration flushInterval,
       required HttpClient client,
--- a/packages/common_client/lib/src/ld_common_client.dart
+++ b/packages/common_client/lib/src/ld_common_client.dart
@@ -139,7 +139,7 @@ final class LDCommonClient {
   // If there are cross-dependent modifiers, then this must be considered.
   late final List<ContextModifier> _modifiers;
 
-  EventProcessorFactory _eventProcessorFactory;
+  final EventProcessorFactory _eventProcessorFactory;
 
   /// The event processor is not constructed during LDCommonClient construction
   /// because it requires the HTTP properties which must be determined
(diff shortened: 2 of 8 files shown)


packages/common/lib/src/events/default_event_processor.dart
```diff
@@ -50,7 +50,7 @@ final class DefaultEventProcessor implements EventProcessor {
 
   DefaultEventProcessor(
       {required LDLogger logger,
-      bool indexEvents = false,
+      required bool indexEvents,
```
tanderson-ld, line 53: For reviewers: I was trying to find a way to not have to tweak this, but the typedef of EventProcessorFactory was not handling the optional named parameter with default well.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -77,6 +77,49 @@ Map<ConnectionMode, DataSourceFactory> _defaultFactories(
...
+    required Duration flushInterval,
+    required HttpClient client,
+    required String analyticsEventsPath,
+    required String diagnosticEventsPath,
+    required ServiceEndpoints endpoints,
+    required Duration diagnosticRecordingInterval,
+    required bool allAttributesPrivate,
+    required Set<AttributeReference> globalPrivateAttributes,
+    DiagnosticsManager? diagnosticsManager}
+    );
```
tanderson-ld, line 93: For reviewers: Added for test injection.

packages/common_client/test/ld_dart_client_test.dart
```diff
@@ -284,4 +285,66 @@ void main() {
...
+                required endpoints,
+                required eventCapacity,
+                required flushInterval,
+                required globalPrivateAttributes,
+                required indexEvents,
+                required logger}) =>
+            mockEventProcessor,
+      );
+    });
+
```
tanderson-ld, line 332: TODO: Add more tests.

packages/common_client/lib/src/ld_common_client.dart
```diff
@@ -528,6 +576,12 @@ final class LDCommonClient {
     LDEvaluationDetail<LDValue> detail;
 
     if (evalResult != null && evalResult.flag != null) {
+      if (evalResult.flag?.prerequisites != null) {
```
kinyoklion, line 579: I think you could remove this check entirely. As it is covered by `?.forEach`

packages/common/lib/src/events/default_event_processor.dart
kinyoklion: I am not sure I see why the type here needs to change, it seems like it would just get redundantly passed, which should be fine. Like required at the factory level, but not at the constructor level.

That said I am indifferent to the change.

kinyoklion (approved):

packages/common_client/lib/src/ld_common_client.dart
tanderson-ld: Done

PR #184, opened 2025-04-15
feat: Add support for WASM.

kinyoklion:


packages/common/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ dependencies:
   http_parser: ^4.0.0
   uuid: ">= 3.0.7 <5.0.0"
   crypto: ^3.0.3
+  web: ^1.1.1
```
kinyoklion, line 16: Replaces `dart:html`

packages/flutter_client_sdk/example/web/index.html
kinyoklion, line 1: This file was regenerated.

packages/event_source_client/lib/src/sse_client_html.dart
```diff
@@ -53,26 +54,26 @@ class HtmlSseClient implements SSEClient {
...
   }
 
-  void _handleMessageEvent(html.Event event) {
+  void _handleMessageEvent(web.Event event) {
     _activeSince = DateTime.now().millisecondsSinceEpoch;
-    final messageEvent = event as html.MessageEvent;
+    final messageEvent = event as web.MessageEvent;
     final ldMessageEvent = ld_message_event.MessageEvent(
-        messageEvent.type, messageEvent.data, messageEvent.lastEventId);
+        messageEvent.type, (messageEvent.data as JSString).toDart, messageEvent.lastEventId);
```
kinyoklion, line 76: This cast should be done safer.

packages/event_source_client/lib/src/sse_client_html.dart
```diff
@@ -53,27 +54,29 @@ class HtmlSseClient implements SSEClient {
...
-    final messageEvent = event as html.MessageEvent;
-    final ldMessageEvent = ld_message_event.MessageEvent(
-        messageEvent.type, messageEvent.data, messageEvent.lastEventId);
-    _messageEventsController.sink.add(ldMessageEvent);
+    final messageEvent = event as web.MessageEvent;
+    if (messageEvent.data != null && messageEvent.data.typeofEquals('string')) {
+      final ldMessageEvent = ld_message_event.MessageEvent(messageEvent.type,
+          (messageEvent.data as JSString).toDart, messageEvent.lastEventId);
+      _messageEventsController.sink.add(ldMessageEvent);
+    }
```
abarker-launchdarkly, line 79: @kinyoklion This is my attempt to make this cast safer.  `ld_message_event.MessageEvent` will not accept null for the data parameter so my thought is that if we can't cast it then we just skip handling the message.  Thoughts?

PR #259, opened 2026-04-27
feat: add Initializer and Synchronizer source contracts 

kinyoklion:
BEGIN_COMMIT_OVERRIDE
feat: add Initializer and Synchronizer source contracts 
feat: add FDv2 requestor and polling base
END_COMMIT_OVERRIDE

  Adds the HTTP layer for FDv2 polling, plus the Initializer and Synchronizer contracts that future polling and streaming sources will implement.                 
                                                                                                                                                                  
  The requestor is a thin HTTP wrapper: builds the URL, sends GET (context in path) or POST (context in body), tracks ETag for if-none-match, and returns a       
  response record. It knows nothing about the FDv2 protocol.                                                                                                      
                                                                                                                                                                  
  The polling base wraps the requestor with FDv2 semantics: classifies HTTP errors as interrupted vs terminal, treats 304 as a no-op change set, detects the      
  x-ld-fd-fallback header, and runs the response body through a fresh protocol handler to produce an FDv2SourceResult.

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Adds new FDv2 polling HTTP/parse pipeline (URL construction, ETag handling, status/error classification) where mistakes could affect data freshness, retry behavior, or leak PII via logs, though changes are largely additive and covered by tests.
> 
> **Overview**
> Introduces initial FDv2 source abstractions (`Initializer`, `Synchronizer`, plus selector/ping typedefs) to standardize one-shot initialization and long-lived update streams.
> 
> Adds an FDv2 polling implementation split into a thin HTTP layer (`FDv2Requestor`) and a protocol-aware wrapper (`FDv2PollingBase`): supports GET (context in path) vs POST (context in body), merges custom base URLs/query params, tracks `ETag`/`if-none-match`, treats `304` as a no-op payload, classifies recoverable vs terminal HTTP errors, parses FDv2 event collections into `FDv2SourceResult`, and annotates results with `x-ld-fd-fallback`/`x-ld-envid` while sanitizing error/log messages to avoid leaking encoded context.
> 
> Adds comprehensive unit tests for endpoint paths, request construction/ETag/query handling, response parsing/error cases, fallback behavior, and log sanitization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eba8fb081e228964dd57f48c374d7fd2a65880c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
```diff
@@ -0,0 +1,22 @@
...
+  /// request body) and as the prefix for GET requests via [streamingGet].
+  static const String streaming = '/sdk/stream/eval';
+
+  /// Builds the polling GET path with the base64url-encoded context
+  /// embedded in the URL path.
+  static String pollingGet(String encodedContext) => '$polling/$encodedContext';
+
+  /// Builds the streaming GET path with the base64url-encoded context
+  /// embedded in the URL path.
+  static String streamingGet(String encodedContext) =>
```
kinyoklion, line 20: In the FDv1 endpoints these methods included the decision about where the credential would go. That now isn't at the endpoints layer. It isn't covered in this PR, but a distinction will need to be made about when to put into a header versus when to use a query parameter. I think this a discriminator on the event source capability in JS.

kinyoklion: We will either add a "capabilities" field to the event source, or have some `hadCapability(SSECapabilities.requestHeaders)` method. So the event source can determine if it uses headers or query parameters.

kinyoklion:
bugbot review

packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
```diff
@@ -0,0 +1,22 @@
+/// FDv2 endpoint paths.
+///
+/// These paths are uniform across mobile and browser SDKs; FDv2 does
+/// not distinguish between platforms at the endpoint level.
+abstract final class FDv2Endpoints {
```
tanderson-ld, line 5: Why is this abstract?

tanderson-ld (approved):

packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
kinyoklion: In this situation it is basically like static, but there isn't a static class modifier.

So this is abstract, and cannot be instantiated, and final so it cannot be inherited from.

PR #261, opened 2026-04-29
feat: add polling and cache sources for FDv2

kinyoklion:
Implements the four building blocks Phase 3 needs from Stream A:

- calculate_poll_delay.dart: pure helper. Given freshness and interval, returns the time remaining in the interval; zero when overdue; full interval when there's no prior freshness or the freshness is in the future (clock skew clamp).
- polling_initializer.dart: one-shot Initializer. Calls the injected PollFunction up to 3 times. ChangeSetResult and terminal status results return immediately. Interrupted results retry after a 1s delay. After 3 interrupted attempts the last is escalated to terminalError so the orchestrator stops retrying at this layer. close() interrupts a pending retry delay and yields shutdown.
- polling_synchronizer.dart: long-lived Synchronizer. Single- subscription StreamController that polls immediately on subscribe, then schedules subsequent polls via calculatePollDelay over the freshness of the most recent successful result. Interrupted results pass through but do not advance freshness, so a transient failure does not delay the catch-up poll. Injected TimerFactory and now() for deterministic tests.
- cache_initializer.dart: Initializer that reads the persistence cache via an injected CachedFlagsReader. Cache hit emits a full ChangeSetResult with persist=false and an empty selector (the cache does not track server-side selector state). Cache miss or reader exception emits a none-type ChangeSetResult so the initializer chain advances. The reader typedef leaves the actual persistence wiring to the orchestrator (Phase C1).

PollFunction, DelayFunction, TimerFactory, and CachedFlagsReader are typedefs rather than concrete dependencies so the orchestrator can wire real implementations and tests can inject scripted ones, mirroring the abstraction style established in SDK-2183.

**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [ ] I have followed the repository's [pull request submission guidelines](../blob/main/CONTRIBUTING.md#submitting-pull-requests)
- [ ] I have validated my changes against all supported platform versions

**Related issues**

Provide links to any issues in this repository or elsewhere relating to this pull request.

Provide a clear and concise description of what you expect to happen.

Provide a clear and concise description of any alternative solutions or features you've considered.

**Additional context**

Add any other context about the pull request here.

---

> [!NOTE]
> **Medium Risk**
> Adds new FDv2 initialization and steady-state polling components (cache load, retry logic, timers) that affect how flag data is sourced and scheduled; errors here could cause delayed updates or startup failures. Risk is moderated by narrow scope and comprehensive unit tests covering shutdown/retry/timing edge cases.
> 
> **Overview**
> Adds new FDv2 data-source building blocks for **startup-from-cache** and **polling-based sync**.
> 
> Introduces `CacheInitializer` to read persisted flag evaluations and emit a full `ChangeSetResult` with `persist: false` (or a `PayloadType.none` miss to advance the initializer chain), plus a pure `calculatePollDelay` helper to schedule polls based on last successful freshness.
> 
> Adds `FDv2PollingInitializer` (one-shot poll with bounded retries on `interrupted`, abortable via `close`) and `FDv2PollingSynchronizer` (single-subscription polling stream that polls immediately, schedules future polls from freshness, and cleanly stops on cancel/close). Comprehensive unit tests are added for cache hit/miss, retry/shutdown behavior, and poll scheduling edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b6e824dcd3608366b4cbfe40e02d2ead7fdb8b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
@@ -1,0 +1,109 @@
+import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart';
+
+import 'flag_eval_mapper.dart';
+import 'payload.dart';
+import 'selector.dart';
+import 'source.dart';
+import 'source_result.dart';
+
+/// The shape of a cache hit: parsed evaluation results plus the
+/// environment ID that was current when the cache was written.
+typedef CachedFlags = ({
+  Map<String, LDEvaluationResult> flags,
+  String? environmentId,
+});
+
+/// Reads cached flag state for [context] from persistence. Returns
+/// null on a cache miss, an unreadable entry, or a parse failure.
+typedef CachedFlagsReader = Future<CachedFlags?> Function(LDContext context);
+
+/// One-shot initializer that brings the SDK up from its persistence
+/// cache. Always returns synchronously (modulo the persistence read);
+/// retries are not meaningful for a local cache.
+///
+/// On cache hit, emits a [ChangeSetResult] with `persist: false` (the
+/// data came from the cache; writing it back is a no-op) and an empty
+/// selector (the cache does not track server-side selector state).
+/// The payload type is [PayloadType.full]: a cache load is a complete
+/// snapshot, not a delta.
+///
+/// On cache miss, emits a [ChangeSetResult] with [PayloadType.none] so
+/// the initializer chain advances rather than terminating. The cache
+/// is best-effort, not a source of truth.
+final class CacheInitializer implements Initializer {
+  final CachedFlagsReader _reader;
+  final LDContext _context;
+  final LDLogger _logger;
+  final DateTime Function() _now;
+
+  bool _closed = false;
+
+  CacheInitializer({
+    required CachedFlagsReader reader,
+    required LDContext context,
+    required LDLogger logger,
+    DateTime Function()? now,
+  })  : _reader = reader,
+        _context = context,
+        _logger = logger.subLogger('CacheInitializer'),
+        _now = now ?? DateTime.now;
+
+  @override
+  Future<FDv2SourceResult> run() async {
+    if (_closed) return _shutdown();
+
+    final CachedFlags? cached;
+    try {
+      cached = await _reader(_context);
+    } catch (err) {
+      _logger.warn('Cache read failed (${err.runtimeType}); '
+          'treating as miss');
+      return _miss();
+    }
+
+    if (_closed) return _shutdown();
+
+    if (cached == null) {
+      return _miss();
+    }
+
+    final updates = <Update>[];
+    cached.flags.forEach((key, evalResult) {
+      updates.add(Update(
+        kind: flagEvalKind,
+        key: key,
+        version: evalResult.version,
+        object: LDEvaluationResultSerialization.toJson(evalResult),
+      ));
+    });
+
+    return ChangeSetResult(
+      payload: Payload(
+        type: PayloadType.full,
+        selector: Selector.empty,
+        updates: updates,
+      ),
+      environmentId: cached.environmentId,
+      freshness: _now(),
+      persist: false,
+    );
+  }
+
+  @override
+  void close() {
+    _closed = true;
+  }
+
+  ChangeSetResult _miss() => ChangeSetResult(
+        payload: const Payload(
+          type: PayloadType.none,
+          updates: [],
+        ),
+        freshness: _now(),
+        persist: false,
+      );
+
+  StatusResult _shutdown() => FDv2SourceResults.shutdown(
+        message: 'Cache initializer closed before completion',
+      );
+}
--- a/packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
@@ -21,9 +21,9 @@ typedef CachedFlagsReader = Future<CachedFlags?> Function(LDContext context);
 /// cache. The cache is read once; retries are not meaningful for a
 /// local read.
 ///
-/// On cache hit, emits a [ChangeSetResult] with `persist: false` (the
-/// data came from the cache; writing it back is a no-op) and an empty
-/// selector (the cache does not track server-side selector state).
+/// On cache hit, emits a [ChangeSetResult] with `persist: false`
+/// (the data is already cached) and an empty selector (the cache
+/// does not track server-side selector state).
 /// The payload type is [PayloadType.full]: a cache load is a complete
 /// snapshot, not a delta.
 ///
--- a/packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
@@ -18,8 +18,8 @@ typedef CachedFlags = ({
 typedef CachedFlagsReader = Future<CachedFlags?> Function(LDContext context);
 
 /// One-shot initializer that brings the SDK up from its persistence
-/// cache. Always returns synchronously (modulo the persistence read);
-/// retries are not meaningful for a local cache.
+/// cache. The cache is read once; retries are not meaningful for a
+/// local read.
 ///
 /// On cache hit, emits a [ChangeSetResult] with `persist: false` (the
 /// data came from the cache; writing it back is a no-op) and an empty
--- a/packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
@@ -1,0 +1,28 @@
+/// Computes how long to wait before the next poll, given when the SDK
+/// last received a fresh response and the configured polling interval.
+///
+/// Returns the time remaining in the interval relative to [freshness].
+/// If [freshness] is null (no successful poll yet) or older than the
+/// interval (we're overdue), returns either the full interval or zero
+/// respectively.
+///
+/// Caps the returned delay at [interval] so a freshness timestamp from
+/// the future (clock skew, manually adjusted system time) cannot push
+/// the next poll arbitrarily far out.
+Duration calculatePollDelay({
+  required DateTime now,
+  required Duration interval,
+  DateTime? freshness,
+}) {
+  if (freshness == null) {
+    return interval;
+  }
+  final elapsed = now.difference(freshness);
+  if (elapsed.isNegative) {
+    return interval;
+  }
+  if (elapsed >= interval) {
+    return Duration.zero;
+  }
+  return interval - elapsed;
+}
--- a/packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
@@ -2,9 +2,8 @@
 /// last received a fresh response and the configured polling interval.
 ///
 /// Returns the time remaining in the interval relative to [freshness].
-/// If [freshness] is null (no successful poll yet) or older than the
-/// interval (we're overdue), returns either the full interval or zero
-/// respectively.
+/// If [freshness] is null (no successful poll yet) returns the full interval.
+/// If [freshness] is older than the interval (we're overdue), returns zero.
 ///
 /// Caps the returned delay at [interval] so a freshness timestamp from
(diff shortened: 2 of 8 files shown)


packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
```diff
@@ -0,0 +1,28 @@
...
+///
+/// Returns the time remaining in the interval relative to [freshness].
+/// If [freshness] is null (no successful poll yet) or older than the
+/// interval (we're overdue), returns either the full interval or zero
+/// respectively.
+///
+/// Caps the returned delay at [interval] so a freshness timestamp from
+/// the future (clock skew, manually adjusted system time) cannot push
+/// the next poll arbitrarily far out.
+Duration calculatePollDelay({
```
kinyoklion, line 11: For now we are always going always poll immediately on a cache initializer. This is used for the synchronizer though.

packages/common_client/lib/src/data_sources/fdv2/polling_initializer.dart
```diff
@@ -0,0 +1,110 @@
...
+/// - [SourceState.interrupted] as transient — retried after
+///   [_retryDelay] up to the attempt limit. After the limit, the last
+///   interrupted result is converted into a [SourceState.terminalError]
+///   so the orchestrator stops retrying at this layer of the chain.
+///
+/// Calling [close] before [run] completes signals an abort: any pending
+/// retry delay returns immediately and [run] resolves to a
+/// [SourceState.shutdown] result.
+final class FDv2PollingInitializer implements Initializer {
+  static const int _maxAttempts = 3;
```
kinyoklion, line 35: I don't think we formalized this into the spec, but for web based/web potential SDKs, I think it makes sense.

packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
```diff
@@ -0,0 +1,109 @@
...
+/// Reads cached flag state for [context] from persistence. Returns
+/// null on a cache miss, an unreadable entry, or a parse failure.
+typedef CachedFlagsReader = Future<CachedFlags?> Function(LDContext context);
+
+/// One-shot initializer that brings the SDK up from its persistence
+/// cache. The cache is read once; retries are not meaningful for a
+/// local read.
+///
+/// On cache hit, emits a [ChangeSetResult] with `persist: false` (the
+/// data came from the cache; writing it back is a no-op) and an empty
```
tanderson-ld: The phrasing of `(the data came from the cache; writing it back is a no-op)` is a little weird.

packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
```diff
@@ -0,0 +1,28 @@
+/// Computes how long to wait before the next poll, given when the SDK
+/// last received a fresh response and the configured polling interval.
+///
+/// Returns the time remaining in the interval relative to [freshness].
+/// If [freshness] is null (no successful poll yet) or older than the
+/// interval (we're overdue), returns either the full interval or zero
+/// respectively.
```
tanderson-ld:
```
/// Returns the time remaining in the interval relative to [freshness].
/// If [freshness] is null (no successful poll yet) returns the full interval.
/// If [freshness] is older than the interval (we're overdue), returns zero.
```

packages/common_client/lib/src/data_sources/fdv2/polling_initializer.dart
```diff
@@ -0,0 +1,110 @@
...
+          'last error: ${lastInterrupted?.message}',
+      statusCode: lastInterrupted?.statusCode,
+      fdv1Fallback: lastInterrupted?.fdv1Fallback ?? false,
+    );
+  }
+
+  @override
+  void close() {
+    if (!_closedSignal.isCompleted) {
+      _closedSignal.complete();
```
tanderson-ld, line 99: Can this just be unconditional?

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
```diff
@@ -0,0 +1,130 @@
...
+    _timer?.cancel();
+    _timer = null;
+    if (!_controller.isClosed) {
+      _controller.add(
+          FDv2SourceResults.shutdown(message: 'Polling synchronizer closed'));
+      _controller.close();
+    }
+  }
+
+  void _onListen() {
```
tanderson-ld, line 79: Can multiple onListens be triggered? I imagine in the internal usage it is never multiple.  Is there a trivial way to make this defensive. If it isn't trivial, no worries.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
```diff
@@ -0,0 +1,130 @@
...
+  final PollFunction _poll;
+  final SelectorGetter _selectorGetter;
+  final Duration _interval;
+  final TimerFactory _timerFactory;
+  final DateTime Function() _now;
+  final LDLogger _logger;
+
+  late final StreamController<FDv2SourceResult> _controller;
+  Timer? _timer;
+  bool _closed = false;
```
tanderson-ld: The polling initilaizer used `Completer<void> _closedSignal = Completer<void>()` for integrating the closed signal.  I would think both the polling initializer and polling synchronizer could use the same mechanism / type for tracking closed.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
```diff
@@ -0,0 +1,130 @@
...
+
+    if (_closed) return;
+
+    if (result is ChangeSetResult && result.freshness != null) {
+      _lastFreshness = result.freshness;
+    }
+
+    if (!_controller.isClosed) {
+      _controller.add(result);
+    }
```
tanderson-ld: I don't think you technically need this isClosed check since you have `if (_closed) return;` just up ahead.  It does feel like there is a bit of double duty of tracking closed-ness with the controller and bool.

I left another comment related to tracking of closed-ness.  Perhaps the controller can be the authority and the closed bool goes away?

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
```diff
@@ -0,0 +1,130 @@
...
+    try {
+      result = await _poll(basis: _selectorGetter());
+    } catch (err) {
+      // PollFunction is the FDv2PollingBase.pollOnce contract, which
+      // already converts errors to StatusResult. A throw here means
+      // someone wired a non-conforming function; treat defensively.
+      _logger.error('Poll function threw unexpectedly: ${err.runtimeType}');
+      if (!_closed && !_controller.isClosed) {
+        _controller.add(FDv2SourceResults.interrupted(
+            message: 'Polling source raised unexpectedly'));
```
tanderson-ld:
```suggestion
            message: 'Polling source raised error unexpectedly'));
```

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
```diff
@@ -0,0 +1,130 @@
...
+    if (_closed) return;
+    final FDv2SourceResult result;
+    try {
+      result = await _poll(basis: _selectorGetter());
+    } catch (err) {
+      // PollFunction is the FDv2PollingBase.pollOnce contract, which
+      // already converts errors to StatusResult. A throw here means
+      // someone wired a non-conforming function; treat defensively.
+      _logger.error('Poll function threw unexpectedly: ${err.runtimeType}');
+      if (!_closed && !_controller.isClosed) {
```
tanderson-ld: Is this condition necessary? If the controller is closed, will add do anything?  I guess it doesn't hurt to have the condition.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
```diff
@@ -0,0 +1,130 @@
...
+      _controller.add(
+          FDv2SourceResults.shutdown(message: 'Polling synchronizer closed'));
+      _controller.close();
+    }
+  }
+
+  void _onListen() {
+    // Kick off the first poll immediately. Subsequent polls are
+    // scheduled from inside _doPoll via the timer.
+    Future<void>.microtask(_doPoll);
```
tanderson-ld: From my very brief reading, this seems like not the intended use of microtask.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: The default `StreamController()` constructor (no `.broadcast()`) creates a single-subscription stream — Dart enforces that you can only `.listen()` to it once. A second `.listen()` call throws `StateError: Stream has already been listened to.` So the runtime guarantees `onListen` fires at most once per controller lifetime; no defensive guard needed at our layer.

packages/common_client/lib/src/data_sources/fdv2/polling_initializer.dart
kinyoklion: The guard is needed to keep `close()` idempotent (per the `Initializer` contract). `Completer.complete()` throws `StateError: Future already completed` if called twice — verified by removing the guard and running the `close is idempotent` test, which then fails with that exact error.

packages/common_client/lib/src/data_sources/fdv2/cache_initializer.dart
kinyoklion: Reworded to "(the data is already cached)" in 6b6e824.

packages/common_client/lib/src/data_sources/fdv2/calculate_poll_delay.dart
kinyoklion: Applied your suggestion verbatim in 6b6e824.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: Reworked the closed-ness tracking in 6b6e824 -- both files now use `Completer<void> _stoppedSignal` as the single authority. See replies below.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: Switched to `Completer<void> _stoppedSignal` in 6b6e824, matching the polling initializer.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: Right -- in 6b6e824 the `_closed` bool is gone and `_controller.isClosed` is no longer checked anywhere. The `Completer<void> _stoppedSignal` is the single authority and it's always completed *before* the controller is closed in `close()`, so the early-return on `_stoppedSignal.isCompleted` is sufficient.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: Suggestion applied in 6b6e824 ("raised error unexpectedly").

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: You're right -- in 6b6e824 the `!_controller.isClosed` half is gone. The `_stoppedSignal.isCompleted` guard alone is sufficient because `close()` completes the signal before closing the controller, so any post-close re-entry into `_doPoll` short-circuits at the signal check.

packages/common_client/lib/src/data_sources/fdv2/polling_synchronizer.dart
kinyoklion: Right -- swapped to `unawaited(_doPoll())` in 6b6e824. The intent was fire-and-forget (since `onListen` is sync and `_doPoll` returns `Future<void>`); `unawaited` from `dart:async` is the canonical way to express that.

tanderson-ld (approved):

PR #263, opened 2026-05-04
fix: Correct conditional-request header and sanitize network error log in FDv1 polling requestor

kinyoklion:
Two FDv1 polling bugs surfaced during review of the FDv2 polling work. The FDv2 sibling already handles both correctly; this PR mirrors that pattern in `packages/common_client/lib/src/data_sources/requestor.dart`.

Jira: [SDK-2322](https://launchdarkly.atlassian.net/browse/SDK-2322)

## 1. Conditional-request header

The requestor was sending the stored ETag back as `etag: <value>`. The conditional-request header per RFC 7232 is `If-None-Match`, not `etag` (which is a *response* header). The server didn't recognize the request as conditional, so 304s were never emitted for our polling traffic and the SDK paid full bandwidth on every successful poll.

If a 304 ever did arrive (non-compliant proxy), the same code path fell through to `DataEvent('put', '', environmentId)` with an empty body, which downstream parsed as empty JSON and surfaced as `ErrorKind.invalidData` — the opposite of the intended "data is current" semantics.

**Fix:** send `if-none-match`, treat 304 as `null` (no change), treat 200 as fresh body.

## 2. Network-error log message

The catch block passed `err.toString()` into both the warn log and the public `StatusEvent.message`. For `http.ClientException`, `toString()` formats as `'ClientException: <msg>, uri=<full-url>'`, so the request URL ended up in both places — and that URL contains the base64url-encoded context.

**Fix:** categorize the exception via minification-safe `is` checks (`TimeoutException`, `http.ClientException`, plus a substring fallback for `dart:io`'s `TlsException` / `HandshakeException`) into a fixed sanitized message, and use that for both the log line and the `StatusEvent`.

## Testing

- ETag round-trip: response with `etag: abc-123` results in `if-none-match: abc-123` on the next request, and the wrong `etag` header is *not* sent.
- 304 path: a 304 response no longer surfaces as `ErrorKind.invalidData` and leaves `FlagManager` untouched.
- Log content: a network-error path's captured log records do not contain the encoded context (verified via mocktail `MockLogAdapter`).

`dart format`, `dart analyze lib test`, and the polling test suite all pass.

[SDK-2322]: https://launchdarkly.atlassian.net/browse/SDK-2322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Adjusts polling HTTP semantics (ETag/304 handling) and the polling loop scheduling, which could change request frequency and cache behavior if incorrect. Mitigated by added regression tests covering header behavior, 304 handling, and log sanitization.
> 
> **Overview**
> Fixes FDv1 polling conditional requests by sending cached ETags as `if-none-match` (instead of `etag`), treating `304 Not Modified` as *no change*, and avoiding storing invalid/empty ETags (while preserving the previous ETag when the header is omitted).
> 
> Hardens polling failure reporting by replacing raw exception `toString()` with a sanitized, categorized message (timeout/network/TLS/unknown) for both logs and `StatusEvent`s, preventing encoded context leakage. Adds regression tests for ETag round-tripping, 304 behavior (including continuing to poll), and sanitized network-error logging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad22a252a596303247d9baea94cba14c7cc9ff2e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/polling_data_source.dart
+++ b/packages/common_client/lib/src/data_sources/polling_data_source.dart
@@ -95,8 +95,11 @@ final class PollingDataSource implements DataSource {
 
     switch (event) {
       case null:
-        // No change.
-        return;
+        // No change (e.g. 304 Not Modified). Fall through to schedule
+        // the next poll -- returning here would leave the timer
+        // unset and silently stop polling after the first unchanged
+        // response.
+        break;
       case DataEvent():
         _eventController.sink.add(event);
       case StatusEvent():
--- a/packages/common_client/lib/src/data_sources/requestor.dart
+++ b/packages/common_client/lib/src/data_sources/requestor.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:http/http.dart' as http;
 import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart';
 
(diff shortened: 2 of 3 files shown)


tanderson-ld:
Is there end to end test coverage in the flutter repo or via contract tests for this case?

tanderson-ld (approved):

PR #267, opened 2026-05-06
feat: FDv2 streaming base, initializer, and synchronizer

kinyoklion:
**Stacked on #266** (SSE capability surface). Mark this PR ready for review only after #266 merges and the base is rebased onto main.

Jira: [SDK-2185](https://launchdarkly.atlassian.net/browse/SDK-2185)

## Scope

Phase B1 from the parallel plan — the streaming half of FDv2 sources, mirroring the polling vertical that landed in SDK-2183 / SDK-2184.

Three new files in `packages/common_client/lib/src/data_sources/fdv2/`:

### `streaming_base.dart` — `FDv2StreamingBase`
Wraps an `SSEClient` with FDv2 protocol semantics. Single-subscription `StreamController<FDv2SourceResult>`:

- `onListen` → builds a fresh `FDv2ProtocolHandler`, subscribes to the SSE client's stream.
- `onCancel` → tears down without emitting shutdown (subscriber-initiated).
- `close()` → emits shutdown then closes; idempotent.

Per event:
- Named SSE events get parsed as JSON, wrapped in an `FDv2Event`, fed to the handler.
- `ActionPayload` → `ChangeSetResult` with `persist: true`.
- `ActionGoodbye` → goodbye `StatusResult`; closes the connection (server told us to disconnect).
- `ActionServerError` / `ActionError` → interrupted `StatusResult`; SSE client's built-in backoff handles reconnect.
- `ActionNone` → no emission.
- Legacy `ping` events → injected `PingHandler` (one-shot poll); result is forwarded.
- `x-ld-fd-fallback: true` on the `OpenEvent` → terminalError with `fdv1Fallback: true`; closes.
- SSE transport error → interrupted; reconnect via SSE client.

Closed-ness tracked via a single `Completer<void> _stoppedSignal` (matches the polling synchronizer pattern from SDK-2184).

### `streaming_initializer.dart` — `FDv2StreamingInitializer`
Implements `Initializer`. Subscribes to the base, completes `run()` with the first emission, then closes the connection. `close()` before the first emission yields a shutdown result.

### `streaming_synchronizer.dart` — `FDv2StreamingSynchronizer`
Implements `Synchronizer`. Thin adapter forwarding the base's stream and `close()`, so the orchestrator can treat polling and streaming uniformly.

## Auth wiring (deferred to orchestrator)

`FDv2StreamingBase` does not build URLs or set headers. The orchestrator (SDK-2186) constructs the `SSEClient`, deciding via `sseClient.hasCapability(SSECapability.requestHeaders)` whether to set the credential in the `Authorization` header or as the `auth` URL query parameter. The streaming source consumes whatever client it's handed.

## Verification

- 25 new tests across the three files; full FDv2 directory now at 185 tests.
- `dart analyze lib test` clean in `common_client/`.
- `melos run analyze` and `melos run test` both pass workspace-wide.

## Follow-up (out of scope)

- The orchestrator (C1 / SDK-2186) is what actually wires `FDv2StreamingBase` into the SDK. This PR delivers the building blocks.

[SDK-2185]: https://launchdarkly.atlassian.net/browse/SDK-2185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> New long-lived streaming path affects how flag payloads are received and persisted; mistakes could cause stale/corrupt updates or mishandled fallback/goodbye, though behavior is heavily tested and mirrors polling patterns.
> 
> **Overview**
> Adds the **FDv2 streaming** path alongside existing polling: `FDv2StreamingBase` wires `SSEClient` to `FDv2ProtocolHandler`, emitting `ChangeSetResult` / status results on a lazy single-subscription stream with `_stoppedSignal`-based shutdown (mirroring polling).
> 
> **Protocol & resilience:** Resets the handler on each `OpenEvent` and after parse failures so partial transfers cannot cross reconnects. Handles `x-ld-fd-fallback` (terminal FDv1 fallback), goodbye (terminal close), malformed JSON as interrupted, and **sanitized** SSE error logging (no URL/context leakage). Legacy **`ping`** events trigger an injected one-shot poll with **coalescing** (one in flight, at most one follow-up).
> 
> `FDv2StreamingSynchronizer` is a thin `Synchronizer` wrapper. **`FDv2ProtocolHandler`** server-error logs now include payload id (or `<unknown>`) and note automatic retry; tests cover streaming lifecycle, ping coalescing, reconnect safety, and log format.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9f185bfc5f30ab0e0a79ea80e75d5914974ffaa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/common_client/lib/src/data_sources/fdv2/streaming_initializer.dart
kinyoklion, line 1: I am undecided if we should support this. If we don't then we can combine the base into the synchronizer and discard it.

Flutter does run on web, but it is inherently a single page app. So I don't think one-shot mode is nearly as important as it is for js.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -74,28 +81,51 @@ final class FDv2StreamingBase {
...
       _handleEvent,
       onError: _handleSseError,
     );
   }
 
+  /// Builds a fresh [FDv2ProtocolHandler]. Called on initial connect
+  /// and on every subsequent [OpenEvent] (SSE auto-reconnect), so a
+  /// partial transfer from the previous connection cannot bleed into
+  /// the new one regardless of whether the server re-sends
+  /// `server-intent` after a Last-Event-ID resumption. Also called
```
kinyoklion: In practice this `last-event-id` part isn't relevant to us.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -0,0 +1,310 @@
...
+///   closed.
+/// - [ActionServerError] / [ActionError] --> interrupted
+///   [StatusResult]; the SSE client's built-in retry handles the
+///   reconnect.
+/// - [ActionNone] --> no emission (waiting for more events).
+///
+/// Legacy `ping` events are routed to the injected [PingHandler] (which
+/// performs a one-shot poll) and the result is forwarded to the
+/// stream. This is the streaming-to-polling bridge for older servers
+/// that pre-date FDv2.
```
tanderson-ld: Probably can remove `This is the streaming-to-polling bridge for older servers that pre-date FDv2.`

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -0,0 +1,310 @@
...
+///
+/// The `x-ld-fd-fallback` header on the initial connection's response
+/// is detected and produces a terminal-error result with
+/// `fdv1Fallback: true`. The connection is closed.
+///
+/// Lifecycle: a single-subscription stream. [results] starts the SSE
+/// connection on subscribe; cancelling the subscription tears it down
+/// without emitting a shutdown. [close] both stops the source and
```
tanderson-ld: > cancelling the subscription tears it down without emitting a shutdown

Could it be more symmetric to emit the shutdown in that case?  Feel free to disregard if not trivial or there is some complication.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -0,0 +1,310 @@
...
+/// emits a shutdown [StatusResult] before closing the stream. Both
+/// paths funnel through a `Completer<void> _stoppedSignal` so async
+/// callbacks short-circuit safely.
+///
+/// `SSEClient.restart` is intentionally not surfaced here. The
+/// orchestrator drives connection lifecycle by tearing down a
+/// streaming source and constructing a fresh one (e.g. on credential
+/// rotation or basis change), not by reconnecting an existing one.
```
tanderson-ld: Is this credential rotation part in the comment real?

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -0,0 +1,310 @@
...
+    _handler = FDv2ProtocolHandler(
+      objProcessors: {flagEvalKind: processFlagEval},
+      logger: _logger,
+    );
+  }
+
+  Future<void> _onCancel() async {
+    if (_stoppedSignal.isCompleted) return;
+    _stoppedSignal.complete();
+    _tearDownConnection();
```
tanderson-ld, line 128: I was poking around with AI and it seems to think the following is necessary in _onCancel after the tear down.

```
  if (!_controller.isClosed) {
    _controller.close();
  }
```

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+
+  Future<void> _handlePing() async {
+    // The FDv2 ping semantic is "go re-poll". A single in-flight poll
+    // already satisfies any number of pings that arrive while it is
+    // running, so drop excess pings rather than spawning concurrent
+    // polls (which would race on emit-order and amplify load on the
+    // polling endpoint).
+    if (_pingInFlight) return;
```
tanderson-ld: Is this case a concern? I think some SDKs implicitly handle this by synchronously working through the queue of incoming ping messages with no midline returns.

T0  Server state = A
T1  ping 1 arrives. _pingInFlight := true. _pingHandler() starts.
T2  Server snapshots state for the in-flight request (= A).
T3  Server state -> B.
T4  ping 2 arrives. _handlePing sees _pingInFlight, drops it.
T5  Response arrives carrying A. _emit(A). _pingInFlight := false.
T6  Steady state: SDK thinks it has the latest. It does not.

If this case is real, perhaps tweak it to be a "pending ping required" sort of flag.  I think this could require triggering another ping after the in flight response is handled, which gets fun. /s

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    if (_stoppedSignal.isCompleted) return;
+
+    switch (action) {
+      case ActionPayload(:final payload):
+        _emit(ChangeSetResult(
+          payload: payload,
+          environmentId: _environmentId,
+          freshness: _now(),
```
tanderson-ld: nit: get _now() before json decode.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
kinyoklion: That sounds like a violation of how streams work. Because the subscription is cancelled, so emitting a shutdown that cannot be handled or to a closed stream.

But maybe the comment could be removed if it causes confusion.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
kinyoklion: I think it is overly ambitious phrasing.

packages/common_client/lib/src/data_sources/fdv2/streaming_initializer.dart
kinyoklion: Removed

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
kinyoklion: removed

tanderson-ld (approved):

PR #278, opened 2026-05-28
feat: register plugin dynamically

abelonogov-ld:
## Summary

Adds runtime plugin registration through `LDClient.registerPlugin`. This API is not supposed to be used by Customers.

- Stores plugin environment metadata during client construction so runtime plugins receive the same SDK, application, and credential metadata as config-time plugins.
- Registers plugin-provided hooks before invoking `plugin.register`, so hooks are active during registration.
- Skips plugin registration if reading `plugin.hooks` fails, while preserving existing safe handling for registration exceptions.

## Test Plan

- Added tests for runtime plugin registration metadata.
- Added tests verifying runtime plugin hooks are active after registration.
- Added coverage that hooks are installed before `plugin.register` runs.
- Added failure coverage for throwing `hooks` getters and throwing `register` methods.
- Ran:

```sh
flutter test test/ld_client_plugin_test.dart


<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Low Risk**
> Additive SDK API with existing safe plugin helpers; no changes to flag evaluation or credential handling beyond reusing stored metadata.
> 
> **Overview**
> Adds **`LDClient.registerPlugin`** so plugins can be attached after the client is constructed (internal API, not for customers). The constructor now **persists `PluginEnvironmentMetadata`** on the client and reuses it for runtime registration, matching config-time plugins (SDK, app, credential).
> 
> **`registerPlugin`** loads hooks via `safeGetPluginHooks`, **installs them before** `Plugin.register`, then calls the existing `safeRegisterPlugins` path. Throwing **`hooks`** skips registration; throwing **`register`** is logged only.
> 
> Example app **iOS/macOS `.gitignore`** entries ignore regenerated **`Podfile`** / **`Podfile.lock`**.
> 
> Tests cover metadata, hook activation, hook-before-register ordering, and failure modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7058027bff5ae4c9750409c491de0a71ea9a206d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### CODE REVIEW

File: packages/common_client/lib/launchdarkly_common_client.dart
Comment (tanderson-ld) [inline packages/common_client/lib/launchdarkly_common_client.dart:1]: Split common client package changes into its own PR.  This will make releasing the version bump more straightforward if common goes in and releases before the flutter package.

File: packages/flutter_client_sdk/example/ios/Podfile
Comment (tanderson-ld) [inline packages/flutter_client_sdk/example/ios/Podfile:1]: If these files don't need to be committed, I recommend removing them.  (Applies to other pod files as well.)

File: packages/flutter_client_sdk/example/ios/Flutter/AppFrameworkInfo.plist [hunk 1]
```diff
@@ -20,7 +20,5 @@
   <string>????</string>
   <key>CFBundleVersion</key>
   <string>1.0</string>
-  <key>MinimumOSVersion</key>
-  <string>12.0</string>
```
tanderson-ld, line 24: Why has this been removed?  Is this ok?

packages/flutter_client_sdk/lib/src/ld_client.dart
```diff
@@ -358,4 +359,35 @@ interface class LDClient {
...
+    for (final hook in hooks) {
+      addHook(hook);
+    }
+
+    safeRegisterPlugins(
+        this, _pluginEnvironmentMetadata, [plugin], _client.logger);
+  }
+
+  /// Returns plugin hooks for runtime registration, or `null` if retrieval failed.
+  List<Hook>? _getRuntimePluginHooks(Plugin plugin) {
```
tanderson-ld: This function should live in common operations.dart next to safeGetHooks.  Seems like it is different than safeGetHooks just in that it handles one plugin and returns null instead of the empty set in the case of an error.

apps/flutter_client_contract_test_service/pubspec.lock
```diff
@@ -388,33 +388,36 @@ packages:
     source: hosted
     version: "6.8.0"
   launchdarkly_common_client:
-    dependency: "direct overridden"
+    dependency: transitive
```
tanderson-ld: Not sure if this is proper.  I think if you do common as its own PR and that then releases, this will then be able to directly point to the common version and this change/diff will go away.

abelonogov-ld: This file is generated, gonna look other changes

packages/flutter_client_sdk/example/ios/Flutter/AppFrameworkInfo.plist
abelonogov-ld: It is example code
Flutter minimum is 13 and it is being written by CocoaPods or SPM during packaging I believe

tanderson-ld (approved):

PR #297, opened 2026-06-10
feat: Add FDv2 synchronizer fallback and recovery conditions

kinyoklion:
## What this adds

Timed conditions that the FDv2 data source orchestrator (a later PR) observes alongside the active synchronizer's results to drive tier transitions:

- **Fallback condition**: starts its timer when the synchronizer reports an interrupted status and cancels it when a change set arrives. If it fires (120 seconds by default), the orchestrator moves to the next available synchronizer. Terminal statuses (shutdown, terminal error, goodbye) do not arm the timer — the orchestrator reacts to those immediately rather than waiting out a fallback period — and repeated interruptions do not extend the deadline; the period counts from the first interruption.
- **Recovery condition**: starts when observed and ignores results. If it fires (300 seconds by default), the orchestrator returns to the primary synchronizer.

`getConditions` selects which conditions apply: none when only one synchronizer is available (nowhere to fall back to), fallback only for the primary, and both for a non-primary synchronizer. `ConditionGroup` merges its members and emits the first to fire.

The timeout defaults match the other client-side FDv2 implementations.

## Why streams instead of futures

The Java and C++ implementations model conditions as one-shot futures, and both had the same leak (fixed in java-core by `c27bf26` / launchdarkly/java-core#163): a future's listeners can never be detached, only released by completion, so an orchestration loop that races a long-lived pending future per result accumulates an irremovable listener garland per change set — unbounded on a healthy primary. Dart futures have the identical property (measured: ~559 B retained per race against a pending future), but Dart also has the primitive those platforms lack: cancellable stream subscriptions.

Each condition therefore exposes a single-subscription `Stream<ConditionType>` that emits **at most once** and then closes (closing without emitting if the condition is closed first). Lifetimes are scoped to the subscription: self-starting timers begin when the stream is listened to, and cancelling the subscription closes the condition and releases its timers. One bounded edge remains by design — informing a never-listened group can arm a fallback timer until its timeout elapses or `close()` is called — and is documented on the API. The orchestrator PR consumes these with one subscription per synchronizer run, closes in a `finally`, and includes a soak test asserting bounded memory across a sustained stream of results.

## Testing

Tests run inside a `package:fake_async` zone, advancing time with `elapse` and asserting timer state through `pendingTimers` — no timer injection in the production code. They cover timer start/cancel behavior for both condition kinds, terminal statuses not arming the fallback timer, the fallback deadline not extending on repeated interruptions, at-most-once emission including when two member timers contend in the same instant, close semantics, subscription cancellation releasing condition and group timers, group merging and inform broadcast, and the `getConditions` selection rules.

SDK-2186

---

> [!NOTE]
> **Medium Risk**
> New timing logic will drive synchronizer failover when wired into the orchestrator; behavior is isolated in new modules with thorough tests and no production integration in this PR.
> 
> **Overview**
> Adds **FDv2 synchronizer tier-transition conditions** for a future orchestrator: timed signals that emit **fallback** (switch to the next synchronizer after sustained interruption) or **recovery** (return to primary after running on a backup).
> 
> **Fallback** arms on `interrupted`, cancels on a change set, ignores terminal statuses, and does not reset the deadline on repeated interruptions (defaults: 120s). **Recovery** starts when the condition is observed and ignores results (default 300s). Conditions expose **single-subscription streams** (at-most-one emit) so subscriptions can be cancelled without the listener-retention issues of racing futures.
> 
> `ConditionGroup` merges members (first fire wins), broadcasts `inform` to all members, and `getConditions` picks an empty group, fallback-only for primary, or fallback+recovery for non-primary when multiple synchronizers exist.
> 
> Adds **`fake_async`** tests for timer behavior, close/cancel semantics, group contention, and `getConditions` rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fc80f62910349fa35e56c4220db8c17b2327c94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/conditions.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/conditions.dart
@@ -1,0 +1,260 @@
+import 'dart:async';
+
+import 'source_result.dart';
+
+/// Default time a synchronizer may remain interrupted before the
+/// orchestrator falls back to the next available synchronizer.
+const Duration defaultFallbackTimeout = Duration(seconds: 120);
+
+/// Default time a non-primary synchronizer runs before the orchestrator
+/// attempts to recover back to the primary synchronizer.
+const Duration defaultRecoveryTimeout = Duration(seconds: 300);
+
+/// The kind of condition that fired, determining the orchestrator's
+/// response.
+enum ConditionType {
+  /// Move to the next available synchronizer.
+  fallback,
+
+  /// Reset to the primary synchronizer.
+  recovery,
+}
+
+/// A timed condition observed alongside the active synchronizer's
+/// results. When the condition fires, its [events] stream emits a
+/// [ConditionType] that the orchestration loop uses to decide what to
+/// do.
+///
+/// Conditions are streams rather than futures so a consumer can detach:
+/// cancelling a stream subscription releases the consumer's listener,
+/// whereas a listener on a never-completing future can never be
+/// removed and would be retained for the condition's whole lifetime.
+abstract interface class Condition {
+  /// Single-subscription stream that emits at most one [ConditionType]
+  /// when the condition fires and then closes. Closes without emitting
+  /// if the condition is closed first.
+  ///
+  /// The condition's lifetime is scoped to the subscription: timers that
+  /// start on their own start when the stream is listened to, and
+  /// cancelling the subscription closes the condition. A condition that
+  /// is informed but never listened to can hold a timer until its
+  /// timeout elapses; call [close] to release it early.
+  Stream<ConditionType> get events;
+
+  /// Inform the condition about a synchronizer result. Some conditions
+  /// use this to start or cancel their timers.
+  void inform(FDv2SourceResult result);
+
+  /// Cancel any pending timers and close [events]. Idempotent.
+  void close();
+}
+
+final class _TimedCondition implements Condition {
+  final Duration _timeout;
+  final ConditionType _type;
+  final void Function(FDv2SourceResult result,
+      {required void Function() start,
+      required void Function() cancel})? _informHandler;
+
+  late final StreamController<ConditionType> _controller =
+      StreamController<ConditionType>(
+    onListen: _onListen,
+    onCancel: close,
+  );
+  Timer? _timer;
+  bool _closed = false;
+
+  _TimedCondition({
+    required Duration timeout,
+    required ConditionType type,
+    void Function(FDv2SourceResult result,
+            {required void Function() start, required void Function() cancel})?
+        informHandler,
+  })  : assert(timeout > Duration.zero),
+        _timeout = timeout,
+        _type = type,
+        _informHandler = informHandler;
+
+  void _onListen() {
+    // Without an inform handler the timer starts as soon as the
+    // condition is observed (recovery behavior). With one, the handler
+    // decides when to start it.
+    if (_informHandler == null) {
+      _startTimer();
+    }
+  }
+
+  void _startTimer() {
+    if (_timer != null || _closed) return;
+    _timer = Timer(_timeout, () {
+      _timer = null;
+      if (_closed) return;
+      _closed = true;
+      // The controller buffers the event if the subscription has not
+      // started yet, so firing before listen is not lost.
+      _controller.add(_type);
+      _controller.close();
+    });
+  }
+
+  void _cancelTimer() {
+    _timer?.cancel();
+    _timer = null;
+  }
+
+  @override
+  Stream<ConditionType> get events => _controller.stream;
+
+  @override
+  void inform(FDv2SourceResult result) {
+    if (_closed) return;
+    _informHandler?.call(result, start: _startTimer, cancel: _cancelTimer);
+  }
+
+  @override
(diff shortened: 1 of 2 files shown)


kinyoklion:
bugbot review

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
```diff
@@ -0,0 +1,260 @@
...
+}
+
+final class _TimedCondition implements Condition {
+  final Duration _timeout;
+  final ConditionType _type;
+  final void Function(FDv2SourceResult result,
+      {required void Function() start,
+      required void Function() cancel})? _informHandler;
+
+  late final StreamController<ConditionType> _controller =
```
tanderson-ld, line 59: Is a stream controller heavy for this?  Is there another single event stream producing type that is lighter weight?

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
```diff
@@ -0,0 +1,260 @@
...
+    // decides when to start it.
+    if (_informHandler == null) {
+      _startTimer();
+    }
+  }
+
+  void _startTimer() {
+    if (_timer != null || _closed) return;
+    _timer = Timer(_timeout, () {
+      _timer = null;
```
tanderson-ld: I don't think it needs to null the timer since it has the _closed bool providing protection.  I fear nulling the timer in various spots in this code (another spot being cancelTimer) could open it up to a path in which a second timer gets created.

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
```diff
@@ -0,0 +1,260 @@
...
+        _timeout = timeout,
+        _type = type,
+        _informHandler = informHandler;
+
+  void _onListen() {
+    // Without an inform handler the timer starts as soon as the
+    // condition is observed (recovery behavior). With one, the handler
+    // decides when to start it.
+    if (_informHandler == null) {
+      _startTimer();
```
tanderson-ld, line 83: Should the timer start be independent of when someone starts listening? (Not much of a practical difference since the condition is immediately hooked up, but who knows what may happen in the future)

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
```diff
@@ -0,0 +1,260 @@
...
+      _subscriptions.add(condition.events.listen((type) {
+        // First member to fire wins; member timers firing in the same
+        // event-loop turn cannot produce a second emission.
+        if (_fired || _closed) return;
+        _fired = true;
+        _controller.add(type);
+        _finish();
+      }));
+    }
+  }
```
tanderson-ld, line 199: Wondering if futures and a future combining operator would be simpler than streams since there is only one signal being emitted.

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
```diff
@@ -0,0 +1,260 @@
...
+  /// that start on their own start when the stream is listened to, and
+  /// cancelling the subscription closes the group and its members. A
+  /// group that is informed but never listened to can hold a timer
+  /// until its timeout elapses; call [close] to release it early.
+  Stream<ConditionType> get events => _controller.stream;
+
+  void _subscribe() {
+    if (_closed) return;
+    if (_conditions.isEmpty) {
+      _controller.close();
```
tanderson-ld: This _controller.close() feels a little off.  Should this "no conditions" path just call finish() to be consistent with other cleanup?

It also seems like maybe not even having special handling for the empty conditions case is necessary, may keep things simpler to remove.

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
kinyoklion: It wouldn't be a problem in practice, but generally speaking you don't want to start before someone starts listening. 

Like when it is easier to start them immediately that makes sense, but if it is easy to start on use, then I think that is better.

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
kinyoklion: This isn't approach isn't susceptible to the memory leak problem with incomplete promises retaining references to completion.

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
kinyoklion: I don't think that a stream controller is going to be substantially different than a future.

packages/common_client/lib/src/data_sources/fdv2/conditions.dart
kinyoklion: Generally speaking I think the opposite, from a timer perspective. Always nulling and checking ensures that you check you have a timer before starting one. In the closed case I am fine with not nulling it. Though normally I would still null it.

tanderson-ld (approved):

PR #299, opened 2026-06-10
feat: Add FDv2 payload application via FlagManager.applyChanges

kinyoklion:
## What this adds

Two pieces the FDv2 data pipeline (a later PR) needs to apply payloads to the flag store correctly. Nothing produces FDv2 payloads for the flag manager yet, so behavior is unchanged.

### Envelope version is authoritative for flag-eval objects

The flag-eval mapper now parses objects with the `put-object` envelope version substituted in, replacing any in-object value. The FDv2 wire format moved versioning onto the envelope; live service objects carry only `flagVersion`:

```json
{"key":"greekTest","kind":"flag-eval","version":420,
 "object":{"flagVersion":7,"value":true,"variation":0,"trackEvents":false}}
```

The previous parse required an in-object `version`, which made every real payload fail with "FDv2 payload contained invalid data." The contract test harness includes an in-object version in its mock data, which is why the suites did not catch it; verified against the live service. This matches the JS implementation, which spreads the object over `version: update.version`.

### `FlagManager.applyChanges`

Applies a changeset from an FDv2 payload, handling all three transfer types in one place — matching the JS implementation's `applyChanges(context, updates, type)` — plumbed through `FlagPersistence`/`FlagUpdater`:

- **full** — replaces all stored flags. This delegates to the existing init path, so it keeps init's semantics: the context becomes the active context (as with an FDv1 put), changes are reported by diffing against the previous state (including deletions), and the cache is always written — replacing the stored flags with an empty set is a change. `environmentId` is applied on this path.
- **partial** — applies the individual updates without per-item version comparison. FDv2 orders data at the payload level, so a lower envelope version must still apply — the existing `upsert` discards it as out-of-order, which is correct for FDv1 patches but wrong here (the harness's "ignores model version" test sends exactly this). Updates are context-checked like `upsert`, a single change event covers the affected keys, and an empty update set changes nothing and skips the cache write.
- **none** — takes no action: no store mutation, no events, no cache write. The payload confirms the SDK is already up to date; freshness and status handling stay with the caller.

Putting the type dispatch inside the flag manager (rather than having the data pipeline choose between `init` and a partial-only method) keeps the full/partial/none semantics in one self-documenting place and prevents a caller from applying a full payload through the merge path by mistake.

## Known pre-existing divergence, intentionally not addressed here

Change detection (`FlagUpdater._hasChanged`) compares evaluation details by **value only** — `LDEvaluationDetail.==` ignores `variationIndex` and `reason` — so a variation- or reason-only change applies to the store without a `FlagsChangedEvent`, despite the event's documentation promising reason changes. This is shared behavior across `init`/`upsert`/`applyChanges` (the JS SDK compares the whole descriptor) and predates this PR; fixing it here would mix a behavior change into otherwise-inert plumbing, so it is left for a separate change.

## Testing

Mapper tests pin the live wire shape (no in-object version parses; envelope version wins over a differing in-object value), alongside the existing put/delete/unknown-kind coverage. Flag updater tests cover partial transfers (no version comparison, a single change event for the changed keys, tombstones, rejection for an inactive context), full transfers (replacement with change events for the differences including deletions, making the context active, setting the environment ID), and a transfer of none leaving the store untouched and emitting nothing. Persistence-level tests assert a partial apply writes through to the cache, a rejected apply does not, an empty partial and a transfer of none skip the write, and a full transfer always writes — even an empty one.

SDK-2186

---

> [!NOTE]
> **Medium Risk**
> Changes core flag-store update semantics for a new FDv2 path (no per-item version on partial applies, full replace behavior); production behavior is unchanged until callers wire it, but mistakes when integrating could cause stale or overwritten flags.
> 
> **Overview**
> Prepares the Dart common client for FDv2 by fixing **flag-eval** parsing and adding a single entry point to apply FDv2 changesets to the flag store (not wired from the data pipeline yet).
> 
> **Flag-eval mapper:** When deserializing `flag-eval` objects, the put-object **envelope `version`** is merged into the JSON passed to `LDEvaluationResultSerialization.fromJson`, overriding any in-object `version`. That matches live FDv2 wire data where objects only carry `flagVersion`, which previously caused parse failures.
> 
> **`FlagManager.applyChanges`:** New API plumbed through `FlagPersistence` and `FlagUpdater` for `PayloadType.full`, `partial`, and `none`. **Full** replaces all flags via the existing `init` path (active context, diff-based change events, optional `environmentId`, cache always written). **Partial** merges updates **without** per-item version checks (unlike `upsert`), rejects inactive contexts, batches change notifications, and skips cache when the update map is empty. **None** is a no-op (store, events, cache unchanged).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 268c45bd451b2ce029cdbd102026ea32a699521f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/flag_eval_mapper.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/flag_eval_mapper.dart
@@ -22,15 +22,22 @@ Map<String, ItemDescriptor> mapUpdatesToItemDescriptors(List<Update> updates) {
       result[update.key] = ItemDescriptor(version: update.version);
     } else if (update.object case final object?) {
       try {
-        final evalResult = LDEvaluationResultSerialization.fromJson(object);
+        final evalResult = LDEvaluationResultSerialization.fromJson({
+          ...object,
+          // The envelope version is authoritative for FDv2 objects. The
+          // flag-eval object itself carries only flagVersion -- there is
+          // no in-object version on the wire -- so the envelope version
+          // becomes the result's version, replacing any present.
+          'version': update.version,
+        });
         result[update.key] = ItemDescriptor(
           version: update.version,
           flag: evalResult,
         );
       } catch (_) {
-        // Per spec 4.1.2.1: treat unparseable flag_eval as a data source
-        // error. Rethrow so the caller can discard the in-progress payload
-        // and reconnect.
+        // Treat an unparseable flag_eval as a data source error. Rethrow
+        // so the caller can discard the in-progress payload and
+        // reconnect.
         rethrow;
       }
     }
--- a/packages/common_client/lib/src/flag_manager/flag_persistence.dart
+++ b/packages/common_client/lib/src/flag_manager/flag_persistence.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 
 import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart';
+import '../data_sources/fdv2/payload.dart';
 import '../item_descriptor.dart';
 import '../persistence/persistence.dart';
 import 'context_index.dart';
@@ -58,6 +59,23 @@ final class FlagPersistence {
       // We only need to store the cache if there was an update.
       // This is executed asynchronously.
       await _storeCache(context);
+      return true;
+    }
+    return false;
+  }
+
+  Future<bool> applyChanges(
+      LDContext context, Map<String, ItemDescriptor> updates, PayloadType type,
+      {String? environmentId}) async {
+    if (_updater.applyChanges(context, updates, type,
+        environmentId: environmentId)) {
+      // A transfer of none, or a partial transfer with no updates, changes
+      // nothing, so the cache write is skipped. A full transfer always
+      // writes; replacing the stored flags with an empty set is a change.
+      if (type == PayloadType.full ||
+          (type == PayloadType.partial && updates.isNotEmpty)) {
+        await _storeCache(context);
+      }
       return true;
     }
     return false;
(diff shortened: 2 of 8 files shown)


kinyoklion:
bugbot review

packages/common_client/lib/src/data_sources/fdv2/flag_eval_mapper.dart
```diff
@@ -22,15 +22,22 @@ Map<String, ItemDescriptor> mapUpdatesToItemDescriptors(List<Update> updates) {
       result[update.key] = ItemDescriptor(version: update.version);
     } else if (update.object case final object?) {
       try {
-        final evalResult = LDEvaluationResultSerialization.fromJson(object);
+        final evalResult = LDEvaluationResultSerialization.fromJson({
+          ...object,
```
kinyoklion, line 26: This is a fix for real payloads.

I think we need to check contract tests, as they pass without this change. But then this would fail on the real service.

packages/common_client/lib/src/flag_manager/flag_persistence.dart
```diff
@@ -63,6 +64,23 @@ final class FlagPersistence {
     return false;
   }
 
+  Future<bool> applyChanges(
```
kinyoklion, line 67: Previously I had considered maybe not needed this, but I think things just are cleaner with it. Then when we drop FDv1 we can remove upsert/init.

joker23 (approved):

PR #304, opened 2026-06-11
fix: Close the final initializer on exhaustion; document the source manager contract

kinyoklion:
## What this changes

Two small behavior hardenings and an explicit statement of the source manager's API contract.

- **Initializer exhaustion closes the final initializer.** Every other transition closes the previous source when the next one is activated, but the exhaustion path returned null with the last initializer still active. When no synchronizer activation follows, that initializer (and its HTTP client) was held until shutdown. A terminal null now leaves no source running.
- **`engageFdv1Fallback()` is a no-op when no FDv1 fallback slot is configured.** Blocking every slot without unblocking a fallback would leave nothing to activate. The orchestration layer checks before engaging; the guard makes the operation safe regardless of the caller.
- **`hasFdv1FallbackConfigured`** replaces `hasFdv1Fallback` to make clear it reports configuration, not engagement state.

## The documented contract

The class docs now state the assumptions the manager operates under, so they are visible at the API surface rather than implicit in the call sites:

- The manager assumes a single, sequential driver. At most one source is active at a time: activating a source closes the previous one, exhaustion closes the final initializer, and `close` closes whatever remains.
- The synchronizer scan cursor doubles as the active-slot pointer. From an activation (`nextAvailableSynchronizer` / `recreateCurrentSynchronizer`) until the next cursor mutation (`resetSynchronizerIndex` / `engageFdv1Fallback`), the cursor identifies the running synchronizer's slot — `isPrimarySynchronizer` and `blockCurrentSynchronizer` read it and are only meaningful inside that window. After mutating the cursor, a new synchronizer must be activated before consulting either. The driver satisfies this by construction: both reads happen strictly between an activation and the run's outcome being decided, and cursor mutations happen only after the outcome is decided.

## Testing

New tests pin the exhaustion close (a terminal null leaves no source running) and the fallback-engagement guard (engaging with no configured FDv1 slot leaves the slot list untouched). Full package suite passes.

SDK-2186

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/source_manager.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/source_manager.dart
@@ -33,6 +33,21 @@ final class SynchronizerSlot {
 
 /// Manages the state of initializers and synchronizers, tracks which
 /// source is active, and handles source transitions for the orchestrator.
+///
+/// The manager assumes a single, sequential driver and relies on these
+/// contracts:
+///
+/// - At most one source is active at a time. Activating a source closes
+///   the previously active one, exhausting the initializer list closes
+///   the final initializer, and [close] closes whatever remains.
+/// - The synchronizer scan cursor doubles as the active-slot pointer:
+///   from an activation ([nextAvailableSynchronizer] or
+///   [recreateCurrentSynchronizer]) until the next cursor mutation
+///   ([resetSynchronizerIndex] or [engageFdv1Fallback]), the cursor
+///   identifies the running synchronizer's slot. [isPrimarySynchronizer]
+///   and [blockCurrentSynchronizer] read the cursor and are only
+///   meaningful inside that window. After mutating the cursor, activate
+///   a new synchronizer before consulting either of them.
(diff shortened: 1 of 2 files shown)


packages/common_client/lib/src/data_sources/fdv2/source_manager.dart
```diff
@@ -140,14 +159,25 @@ final class SourceManager {
...
 
-  /// Block all non-FDv1 synchronizers and unblock FDv1 synchronizers.
+  /// Block all non-FDv1 synchronizers, unblock the FDv1 fallback, and
+  /// reset the scan cursor so the next activation selects the fallback.
+  /// Does nothing when no FDv1 fallback slot is configured, since
+  /// blocking every slot without unblocking one would leave nothing to
+  /// activate. The active source keeps running until the next
+  /// activation closes it.
   void engageFdv1Fallback() {
+    if (!hasFdv1FallbackConfigured) {
```
kinyoklion, line 178: In practice we may not make this possible.

joker23 (approved):

PR #306, opened 2026-06-12
feat: Add FDv2 streaming source factories and query-parameter authentication

kinyoklion:
> Stacked on #309 (FDv2 source-layer translation); that lands first.

## What this adds

The streaming half of the FDv2 source factories, plus the authentication plumbing both source kinds share. Nothing constructs these factories in production yet (the orchestrator consumes them in a later PR), so behavior is unchanged.

### Streaming factories

`createSynchronizerFactoryFromEntry` now builds `FDv2StreamingSynchronizer`s instead of throwing `UnsupportedError`. The SSE client is constructed with a `uriProvider` that is re-invoked on every connection attempt, so the `basis` query parameter always reflects the current selector — a reconnect after payloads have been applied resumes with a delta request rather than refetching the full payload. The URI builder composes against the configured base URI the same way the polling requestor does (path joining without clobbering existing query parameters).

### Query-parameter authentication

`SourceFactoryContext` gains `additionalQueryParameters`, applied to every data acquisition request by both FDv2 transports — streaming through the `uriProvider`, and polling through `FDv2Requestor`, which merges them into each poll URL. The transports are agnostic about what the parameters carry; authentication enters at the single population site, where `SourceFactoryContext.fromClientConfig` supplies the platform `CredentialConfig.authQueryParameters`: empty on io (every io transport authenticates with the authorization header in the base headers) and `auth=<client-side-id>` on web.

The web platform authenticates with the query parameter at all times rather than the authorization header: a browser only delivers that header when the target service's CORS pre-flight allows it, the browser's native `EventSource` cannot send custom headers at all, and the FDv2 endpoint paths do not embed the credential the way the FDv1 client-side paths do. (Requirement 3.1.3 of the client-side FDv2 spec, added in launchdarkly/sdk-specs#233, permits this.) Verified against the live service: the FDv2 polling endpoint returns 200 with `?auth=<credential>` and 400 without.

## Testing

Factory tests cover the streaming construction path and pin the URI contract: the streaming URI carries the auth query parameters and picks up the current selector's `basis` on each provider invocation, and the polling request URL carries the auth parameters through the factory wiring. Requestor tests assert the additional query parameters appear on every poll URL alongside `basis`, and that the requestor adds no authorization header (authentication comes from base headers or query parameters, never per-request). Full package suite passes.

SDK-2186

---

> [!NOTE]
> **Medium Risk**
> Touches authentication query parameters and streaming connection/error handling; production behavior is unchanged until the orchestrator wires these factories, but mistakes could break web auth or fallback when enabled.
> 
> **Overview**
> Adds **FDv2 streaming synchronizer** factory wiring so `StreamingSynchronizer` entries build `FDv2StreamingSynchronizer` with an SSE client whose **`uriProvider` is rebuilt on each connect**, keeping `basis` aligned with the current selector for delta resumes after reconnect.
> 
> Introduces shared **`buildFDv2Uri`** for polling and streaming so paths, `withReasons`, `basis`, relay-style base query strings (including **duplicate keys**), and extra params merge the same way. **`SourceFactoryContext`** now carries **`credential`** and **`additionalQueryParameters`** (from `CredentialConfig.authQueryParameters`, e.g. web `auth=`), applied on every poll and stream URL via **`FDv2Requestor`** and the streaming URI builder.
> 
> **`FDv2StreamingBase`** gains optional **`defaultEnvironmentId`** when headers are unavailable, and treats **`UnrecoverableStatusError`** as a terminal failure (with optional **`x-ld-fd-fd-fallback`**). Factory and requestor tests cover auth query params, basis on reconnect, and repeated relay query keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 187332ded15c4e930a4c7177a8ebfa1cf3f2ce14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/requestor.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/requestor.dart
@@ -1,5 +1,7 @@
-import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart';
+import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart'
+    hide ServiceEndpoints;
 
+import '../../config/service_endpoints.dart';
 import 'endpoints.dart';
 import 'selector.dart';
 
@@ -39,6 +41,7 @@ final class FDv2Requestor {
   final String _contextJson;
   final bool _usePost;
   final bool _withReasons;
+  final Map<String, String> _additionalQueryParameters;
   String? _lastEtag;
 
   FDv2Requestor({
@@ -49,6 +52,7 @@ final class FDv2Requestor {
     required bool usePost,
     required bool withReasons,
     required HttpProperties httpProperties,
+    Map<String, String> additionalQueryParameters = const {},
     HttpClientFactory httpClientFactory = _defaultHttpClientFactory,
   })  : _logger = logger.subLogger('FDv2Requestor'),
         _baseUri = Uri.parse(endpoints.polling),
@@ -56,6 +60,7 @@ final class FDv2Requestor {
         _contextJson = contextJson,
         _usePost = usePost,
         _withReasons = withReasons,
+        _additionalQueryParameters = additionalQueryParameters,
         _client = httpClientFactory(usePost
             ? httpProperties.withHeaders({'content-type': 'application/json'})
             : httpProperties);
@@ -111,30 +116,12 @@ final class FDv2Requestor {
     final addedPath = _usePost
         ? FDv2Endpoints.polling
         : FDv2Endpoints.pollingGet(_contextEncoded);
-
-    // Compose against the parsed base URI so a custom polling URL
-    // carrying its own query parameters (e.g. a relay proxy with a token)
-    // is preserved correctly. String concatenation against `_baseUri`
-    // would land the appended path inside the query component.
-    final basePath = _baseUri.path.endsWith('/')
-        ? _baseUri.path.substring(0, _baseUri.path.length - 1)
-        : _baseUri.path;
-    final mergedPath = '$basePath$addedPath';
-
-    // Use queryParametersAll so a base URL like `?dup=1&dup=2` round-trips
-    // both values; the simpler `queryParameters` map collapses duplicates.
-    final mergedQuery = <String, dynamic>{};
-    mergedQuery.addAll(_baseUri.queryParametersAll);
-    if (_withReasons) {
-      mergedQuery['withReasons'] = 'true';
-    }
-    if (basis.state case final state? when state.isNotEmpty) {
-      mergedQuery['basis'] = state;
-    }
-
-    return _baseUri.replace(
-      path: mergedPath,
-      queryParameters: mergedQuery.isEmpty ? null : mergedQuery,
+    return buildFDv2Uri(
+      baseUri: _baseUri,
+      addedPath: addedPath,
+      withReasons: _withReasons,
+      basis: basis,
+      additionalQueryParameters: _additionalQueryParameters,
     );
   }
 }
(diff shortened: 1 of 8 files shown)


packages/common_client/lib/src/data_sources/fdv2/requestor.dart
kinyoklion, line 1: After some discussion and consideration I decided that the auth for the browser may as well always use the query parameter. This reduces the risk of a cors problem.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -293,6 +295,24 @@ final class FDv2StreamingBase {
...
+      // orchestrator moves to another source. The error response may also
+      // carry the FDv1 fallback directive.
+      final fallback = err.headers['x-ld-fd-fallback']?.toLowerCase() == 'true';
+      _logger.warn(
+          'Streaming request failed with status ${err.statusCode}; giving up');
+      _terminate(
+          finalResult: FDv2SourceResults.terminalError(
+        message: 'Streaming request failed with status ${err.statusCode}',
+        statusCode: err.statusCode,
+        fdv1Fallback: fallback,
```
kinyoklion, line 332: I don't know if there is strictly any way for this to be true that it is non-recoverable, but it included the fallback. But they are headers that could be attached to any response.

packages/common_client/lib/src/data_sources/fdv2/source_factory_context.dart
```diff
@@ -62,6 +74,13 @@ final class SourceFactoryContext {
       withReasons: withReasons,
       defaultPollingInterval: defaultPollingInterval,
       cachedFlagsReader: cachedFlagsReader,
+      credential: credential,
+      // Platforms that authenticate with the `auth` query parameter
```
joker23, line 78: should we enforce this as a rule?

kinyoklion: I am not sure I understand the question.

joker23 (approved):

packages/common_client/lib/src/data_sources/fdv2/source_factory_context.dart
joker23: resolved in slack

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -314,6 +316,24 @@ final class FDv2StreamingBase {
 
   void _handleSseError(Object err, StackTrace stack) {
     if (_stoppedSignal.isCompleted) return;
+
+    if (err is UnrecoverableStatusError) {
+      // The SSE client stops reconnecting for these status codes, so the
+      // source cannot recover on its own. Surface a terminal error so the
+      // orchestrator moves to another source. The error response may also
+      // carry the FDv1 fallback directive.
+      final fallback = err.headers['x-ld-fd-fallback']?.toLowerCase() == 'true';
```
tanderson-ld, line 325: Seeing fallback be conditional on `err is UnrecoverableStatusError` makes we think it can be pulled up

tanderson-ld (approved):

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
kinyoklion: Headers are only available inside the UnrecoverableStatusError type. The base error type doesn't have them and can represent a runtime error.

PR #309, opened 2026-06-15
feat: Translate FDv2 payloads at the data source layer

kinyoklion:
BEGIN_COMMIT_OVERRIDE
feat: Translate FDv2 payloads at the data source layer
feat: Add FDv2 streaming source factories and query-parameter authentication
END_COMMIT_OVERRIDE

## What this changes

Moves flag-eval translation out of the apply path and into the data source layer, and introduces a typed `ChangeSet` distinct from the wire-level `Payload`.

Previously the streaming and polling sources emitted a `ChangeSetResult` carrying the raw wire `Payload` (a list of `Update`s with unparsed object maps), and the flag-eval objects were converted to typed descriptors later, at apply time in `DataSourceEventHandler.handlePayload`. A payload that parsed at the protocol level but whose flag objects were malformed therefore failed *after* acquisition — surfacing as an invalid-message that drove a connection restart, on a fixed interval, without ever arming the orchestrator's fallback timer.

Now:

- `ChangeSet` carries typed `Map<String, ItemDescriptor>` updates plus the type and selector. `translatePayload` converts a wire `Payload` into a `ChangeSet`, throwing if any flag-eval object cannot be parsed.
- The streaming and polling sources translate at acquisition. A translation failure becomes an **interrupted** source result, exactly like a malformed protocol body — so the orchestrator's fallback timer governs recovery (a source stuck on invalid data falls back after the timeout instead of retrying forever). Streaming discards the partial handler state and keeps the SSE connection; the server's next valid payload is processed normally.
- `ChangeSetResult` carries the `ChangeSet`, and `handlePayload` applies it directly with no conversion and no failure path of its own.
- The cache initializer builds its `ChangeSet` straight from the already-typed cached evaluation results, dropping a JSON round-trip that previously serialized typed results only to re-parse them.

This is foundational for the FDv2 data system: the streaming-source and orchestrator PRs build on `ChangeSet`. Nothing constructs the FDv2 sources in production yet, so behavior is unchanged for shipping code.

## Testing

Source-layer tests cover a protocol-valid payload whose flag data cannot be parsed: polling returns interrupted, and streaming returns interrupted then recovers on the next valid payload. The flag manager / event handler tests apply typed change sets directly (full replace, partial without per-item version comparison, none). Cache initializer tests assert the evaluation result is carried through without re-parsing. Full `common_client` suite passes.

SDK-2186

---

> [!NOTE]
> **Medium Risk**
> Touches FDv2 acquisition, URL/auth composition, and error-classification paths that govern fallback and flag updates; changes are largely behind not-yet-production FDv2 wiring but alter recovery semantics when invalid flag data arrives.
> 
> **Overview**
> FDv2 now separates wire **`Payload`** from typed **`ChangeSet`** (`Map<String, ItemDescriptor>`). **`translatePayload`** runs in polling/streaming (and cache builds descriptors directly), so **`ChangeSetResult`** and **`handlePayload`** only apply already-typed updates.
> 
> **Malformed flag-eval data** is surfaced as **`interrupted`** at the source layer (like other transient parse errors) instead of failing later in the event handler—so the orchestrator’s fallback timer can kick in rather than endless connection retries.
> 
> Adds shared **`buildFDv2Uri`** (relay query params, `basis`, `withReasons`) for polling and streaming; **`SourceFactoryContext`** carries credential-driven **`auth`** query params. **Streaming synchronizer factories** are implemented (SSE client, per-reconnect URI/`basis`, legacy **`ping`** → one-shot poll, env ID fallback). Streaming also maps **`UnrecoverableStatusError`** to terminal errors with optional FDv1 fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38646ce5dd8d599079ccb69f9ded9fa5f5f90dda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
@@ -1,3 +1,7 @@
+import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart';
+
+import 'selector.dart';
+
 /// FDv2 endpoint paths.
 ///
 /// These paths are uniform across mobile and browser SDKs; FDv2 does
@@ -19,4 +23,40 @@ abstract final class FDv2Endpoints {
   /// embedded in the URL path.
   static String streamingGet(String encodedContext) =>
       '$streaming/$encodedContext';
+}
+
+/// Builds an FDv2 request URI: appends [addedPath] to [baseUri]'s path and
+/// merges the [withReasons], [basis], and [additionalQueryParameters]
+/// query parameters onto the base URL's own.
+///
+/// Composes against the parsed [baseUri] so a custom URL carrying its own
+/// query parameters (e.g. a relay proxy with a token) is preserved --
+/// including repeated keys, via `queryParametersAll`, which a plain
+/// `queryParameters` map would collapse to the last value. String
+/// concatenation against the base would instead land the appended path
+/// inside the query component.
+///
+/// Shared by the polling requestor and the streaming source so the two
+/// transports build URLs identically.
+Uri buildFDv2Uri({
+  required Uri baseUri,
+  required String addedPath,
+  required bool withReasons,
+  required Selector basis,
+  Map<String, String> additionalQueryParameters = const {},
+}) {
+  final mergedPath = appendPath(baseUri.path, addedPath);
+  final mergedQuery = <String, dynamic>{}
+    ..addAll(baseUri.queryParametersAll)
+    ..addAll(additionalQueryParameters);
+  if (withReasons) {
+    mergedQuery['withReasons'] = 'true';
+  }
+  if (basis.state case final state? when state.isNotEmpty) {
+    mergedQuery['basis'] = state;
+  }
+  return baseUri.replace(
+    path: mergedPath,
+    queryParameters: mergedQuery.isEmpty ? null : mergedQuery,
+  );
 }
--- a/packages/common_client/lib/src/data_sources/fdv2/payload.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/payload.dart
@@ -1,5 +1,6 @@
 import 'package:launchdarkly_dart_common/launchdarkly_dart_common.dart';
 
+import '../../item_descriptor.dart';
 import 'selector.dart';
 
 /// The type of payload transfer.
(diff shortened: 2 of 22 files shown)


packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
```diff
@@ -20,3 +24,39 @@ abstract final class FDv2Endpoints {
   static String streamingGet(String encodedContext) =>
       '$streaming/$encodedContext';
 }
+
+/// Builds an FDv2 request URI: appends [addedPath] to [baseUri]'s path and
```
kinyoklion, line 28: This was part of 306. Which I rebased on to 309, and then merged into 309. So some of this will be familiar.

packages/common_client/lib/src/data_sources/fdv2/endpoints.dart
```diff
@@ -20,3 +24,39 @@ abstract final class FDv2Endpoints {
...
+/// Composes against the parsed [baseUri] so a custom URL carrying its own
+/// query parameters (e.g. a relay proxy with a token) is preserved --
+/// including repeated keys, via `queryParametersAll`, which a plain
+/// `queryParameters` map would collapse to the last value. String
+/// concatenation against the base would instead land the appended path
+/// inside the query component.
+///
+/// Shared by the polling requestor and the streaming source so the two
+/// transports build URLs identically.
+Uri buildFDv2Uri({
```
kinyoklion, line 41: This is just a helper to make sure we handle the query parameters consistently for streaming and polling.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -293,6 +316,24 @@ final class FDv2StreamingBase {
 
   void _handleSseError(Object err, StackTrace stack) {
     if (_stoppedSignal.isCompleted) return;
+
+    if (err is UnrecoverableStatusError) {
```
kinyoklion, line 320: Same as PR 306.

packages/common_client/lib/src/data_sources/fdv2/payload.dart
```diff
@@ -95,3 +96,36 @@ final class Payload {
...
+/// A [Payload] translated into typed flag descriptors, ready to apply to
+/// the flag store.
+///
+/// The wire-level [Payload] carries raw [Update] objects; producing a
+/// [ChangeSet] converts each flag-eval object into an [ItemDescriptor].
+/// That conversion is performed by the data source layer (see
+/// `translatePayload`) so a malformed object is reported as a data source
+/// error at acquisition time -- where the connection can recover -- rather
+/// than surfacing later, at apply time.
+final class ChangeSet {
```
kinyoklion, line 108: This now matches the ChangeSet type in other SDKs.

tanderson-ld (approved):

PR #310, opened 2026-06-17
feat: Add the FDv2 data system and expose it through configuration

kinyoklion:
## What this adds

The piece that ties the FDv2 building blocks together and turns them on: `FDv2DataSystem`, the manager routing that applies FDv2 payloads, and the configuration that opts in. This completes the FDv2 work in `common_client` — everything below the public Flutter API.

- **`FDv2DataSystem`** composes the data source factories the `DataSourceManager` consumes. It owns the state that must outlive any single orchestrator: the current selector and the context it belongs to. A fresh orchestrator is built per connection-mode switch and per identify; the selector survives mode switches (initializers are skipped when a selector is already held) but resets when the context changes, since a selector is specific to one context. `buildFactories()` returns factories for streaming, polling, and background — offline has no data source and is handled by the manager directly. Custom connection modes from config replace the built-in definitions by name.
- **`DataSourceManager` payload routing.** The `PayloadEvent` case — a no-op placeholder in the orchestrator PR — now applies the change set through `handlePayload` and completes the pending identify, the same way `DataEvent` does for FDv1.
- **Configuration & exposure.** `DataSystemConfig` (providing it, even empty, opts into FDv2); `LDCommonConfig.dataSystem`; `LDCommonClient` constructs an `FDv2DataSystem` and installs its factories when `dataSystem` is configured, otherwise the FDv1 sources run unchanged; and the new public types are exported.

When `dataSystem` is absent the FDv1 path is byte-for-byte unchanged, so existing behavior is unaffected.

## Testing

- `DataSourceManager` test: a `PayloadEvent` is applied through `handlePayload`, marks the source valid, and completes identify (a dropped/no-op payload would leave identify hanging — this distinguishes the real routing from the placeholder).
- `FDv2DataSystem` tests: `buildFactories` exposes streaming/polling/background and not offline; each factory builds a fresh data source per call; a custom connection mode replaces the built-in definition.
- `DataSystemConfig` default (no custom modes).
- Full `common_client` suite passes; `flutter_client_sdk` (the dependent package) analyzes and tests clean against these changes.
- End-to-end behavior of the wired data system is exercised by the FDv2 contract tests, which land with the contract-service changes in a later PR (and pass on the integration branch today).

SDK-2186

---

> [!NOTE]
> **High Risk**
> Touches core flag acquisition, identify completion, connection-mode/offline behavior, and persistence read paths; incorrect timing could leave identifies hanging or serve stale flags, though FDv1 remains unchanged when `dataSystem` is unset.
> 
> **Overview**
> Opts the SDK into **FDv2** when `LDCommonConfig.dataSystem` is set (even empty), exports `DataSystemConfig` / `ConnectionModeId` and mode-definition types, and keeps the FDv1 path when `dataSystem` is absent.
> 
> **`FDv2DataSystem`** builds per-mode orchestrator factories (streaming, polling, background, **offline** as a real cache-only pipeline), holds the **selector** across mode switches, and applies `connectionModes` overrides. **`FDv2DataManager`** clears the selector on each identify and maps `waitForNetworkResults` to cached vs fresh availability; FDv1 still loads cache at identify via **`FDv1DataManager`**.
> 
> **`DataSourceManager`** applies **`PayloadEvent`** through `handlePayload`, sets **valid** on applied online payloads (including no-change restores), does not promote valid in offline mode, and completes identify on first applied data (cached) or **`InitializedEvent`** (fresh). **`InitializedEvent`** and orchestrator init semantics separate cache hits from network-ready state. Cache loads use **`readCached`** without touching the store until the pipeline applies payloads.
> 
> Adds an **FDv1 fallback** synchronizer (FDv1 poll → FDv2 change sets) with loop guards when already on fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 194eae78bb44946888b09b7fc9ab5b5dc6d95d44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/config/data_system_config.dart
+++ b/packages/common_client/lib/src/config/data_system_config.dart
@@ -29,7 +29,7 @@ import '../data_sources/fdv2/mode_definition.dart';
 
 /// Identifies a built-in connection mode whose data-source pipeline can be
 /// overridden through [DataSystemConfig.connectionModes]: [streaming],
-/// [polling], or [background].
+/// [polling], [background], or [offline].
 sealed class ConnectionModeId {
   const ConnectionModeId();
 
@@ -42,6 +42,11 @@ sealed class ConnectionModeId {
   /// The built-in background mode.
   static const ConnectionModeId background =
       _BuiltInConnectionMode('background');
+
+  /// The built-in offline mode. Its pipeline loads cached flags and runs
+  /// no synchronizer, so overriding it customizes how the SDK behaves
+  /// while offline (for example, the cache initializer it uses).
+  static const ConnectionModeId offline = _BuiltInConnectionMode('offline');
 }
 
 final class _BuiltInConnectionMode extends ConnectionModeId {
--- a/packages/common_client/lib/src/config/data_system_config.dart
+++ b/packages/common_client/lib/src/config/data_system_config.dart
@@ -1,0 +1,74 @@
+import '../data_sources/fdv2/mode_definition.dart';
+
+// Maintainer note (not public API): ConnectionModeId is a sealed
+// hierarchy rather than an enum so a custom-mode variant can be added
+// later without changing this surface. The planned extension is a custom
+// variant constructed as `ConnectionModeId.custom('my-mode')`:
+//
+//   factory ConnectionModeId.custom(String name) = _CustomConnectionMode;
+//   final class _CustomConnectionMode extends ConnectionModeId {
+//     final String name;
+//     const _CustomConnectionMode(this.name);
+//     // value equality on name so it works as an override-map key
+//   }
+//
+// A custom mode is a distinct type from a built-in, so the two share no
+// namespace: a custom id never equals a built-in id (even with the same
+// name), and so cannot collide with a current or future built-in. The
+// type is the namespace -- no name prefix is needed. This holds only
+// while custom modes stay typed; if one is ever reduced to a bare string
+// (logs, persistence) that reintroduces a shared string space where a
+// prefix would matter again.
+//
+// Equality split: the built-in values are const singletons relying on
+// canonical-instance identity, which lets a connectionModes map of only
+// built-in keys be a const map. A runtime-constructed custom variant must
+// carry value equality, so an override map holding a custom key would be
+// non-const. The built-in variant therefore must not override
+// `==`/`hashCode`.
+
+/// Identifies a built-in connection mode whose data-source pipeline can be
+/// overridden through [DataSystemConfig.connectionModes]: [streaming],
+/// [polling], or [background].
+sealed class ConnectionModeId {
+  const ConnectionModeId();
+
+  /// The built-in streaming mode.
+  static const ConnectionModeId streaming = _BuiltInConnectionMode('streaming');
+
+  /// The built-in polling mode.
+  static const ConnectionModeId polling = _BuiltInConnectionMode('polling');
+
+  /// The built-in background mode.
+  static const ConnectionModeId background =
+      _BuiltInConnectionMode('background');
+}
+
+final class _BuiltInConnectionMode extends ConnectionModeId {
+  final String name;
+
+  const _BuiltInConnectionMode(this.name);
+
+  @override
+  String toString() => 'ConnectionModeId.$name';
+}
+
+/// Configuration for the FDv2 data system.
+///
+/// Providing a [DataSystemConfig] (even an empty one) opts the SDK into
+/// the FDv2 data acquisition protocol. When absent the SDK uses the
+/// FDv1 data sources.
+///
+/// This feature is not stable, and not subject to any backwards
+/// compatibility guarantees or semantic versioning. It is in early
+/// access. If you want access to this feature please join the EAP.
+final class DataSystemConfig {
+  /// Overrides for built-in connection modes. A definition given here
+  /// replaces the built-in pipeline for that mode; modes not present keep
+  /// their built-in definition.
+  final Map<ConnectionModeId, ModeDefinition> connectionModes;
+
+  const DataSystemConfig({
+    this.connectionModes = const {},
+  });
+}
--- a/packages/common_client/lib/src/data_sources/data_manager.dart
+++ b/packages/common_client/lib/src/data_sources/data_manager.dart
@@ -72,7 +72,9 @@ final class FDv2DataManager implements DataManager {
     _clearSelector();
     final completer = Completer<void>();
     _dataSourceManager.identify(context, completer,
-        requireFreshData: waitForNetworkResults);
+        minimumDataAvailability: waitForNetworkResults
+            ? DataAvailability.fresh
+            : DataAvailability.cached);
     return completer.future;
   }
 }
--- a/packages/common_client/lib/src/data_sources/data_manager.dart
+++ b/packages/common_client/lib/src/data_sources/data_manager.dart
@@ -52,22 +52,22 @@ final class FDv1DataManager implements DataManager {
 /// The cache is not loaded at identify time; the data source pipeline's
 /// cache initializer loads it as the first tier. Identify resolves on the
 /// first delivered payload, or -- when waiting for network results -- only
-/// on basis (network or terminal) data, so a cache load alone does not
-/// satisfy a wait-for-network identify.
+/// on fresh data, so a cache load alone does not satisfy a wait-for-network
+/// identify.
 ///
-/// Identify is also where a held basis is discarded: a selector points at
-/// one context's data, so on a context change [resetBasis] is invoked
+/// Identify is also where a held selector is discarded: a selector points
+/// at one context's data, so on a context change [clearSelector] is invoked
 /// before connecting. This is keyed on the context's canonical key and
 /// driven here rather than inferred from the context instance inside the
 /// data source factory, so it holds regardless of which connection mode is
 /// active when the context changes (including offline).
 final class FDv2DataManager implements DataManager {
   final DataSourceManager _dataSourceManager;
-  final void Function() _resetBasis;
+  final void Function() _clearSelector;
 
   String? _lastContextKey;
 
-  FDv2DataManager(this._dataSourceManager, this._resetBasis);
+  FDv2DataManager(this._dataSourceManager, this._clearSelector);
 
   @override
   Future<void> identify(LDContext context,
@@ -75,7 +75,7 @@ final class FDv2DataManager implements DataManager {
     final key = context.canonicalKey;
     if (key != _lastContextKey) {
       _lastContextKey = key;
-      _resetBasis();
+      _clearSelector();
     }
     final completer = Completer<void>();
     _dataSourceManager.identify(context, completer,
--- a/packages/common_client/lib/src/data_sources/data_manager.dart
+++ b/packages/common_client/lib/src/data_sources/data_manager.dart
@@ -22,12 +22,6 @@ abstract interface class DataManager {
       {required bool waitForNetworkResults});
 }
 
-/// FDv1 data manager.
-///
-/// The cache is loaded imperatively at identify time via
-/// [FlagManager.loadCached]. A cache hit resolves identify immediately
-/// unless the caller is waiting for network results; either way the
-/// network connection is started so live data follows.
 final class FDv1DataManager implements DataManager {
   final DataSourceManager _dataSourceManager;
   final FlagManager _flagManager;
--- a/packages/common_client/lib/src/data_sources/data_manager.dart
+++ b/packages/common_client/lib/src/data_sources/data_manager.dart
@@ -55,28 +55,21 @@ final class FDv1DataManager implements DataManager {
 /// on fresh data, so a cache load alone does not satisfy a wait-for-network
 /// identify.
 ///
-/// Identify is also where a held selector is discarded: a selector points
-/// at one context's data, so on a context change [clearSelector] is invoked
-/// before connecting. This is keyed on the context's canonical key and
-/// driven here rather than inferred from the context instance inside the
-/// data source factory, so it holds regardless of which connection mode is
-/// active when the context changes (including offline).
+/// Each identify starts data acquisition fresh: any held selector is
+/// discarded via [clearSelector] before connecting, so the new connection
+/// re-fetches a full payload rather than resuming a previous context's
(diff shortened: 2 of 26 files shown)


packages/common_client/lib/src/config/data_system_config.dart
```diff
@@ -0,0 +1,21 @@
...
+/// FDv1 data sources.
+///
+/// This feature is not stable, and not subject to any backwards
+/// compatibility guarantees or semantic versioning. It is in early
+/// access. If you want access to this feature please join the EAP.
+final class DataSystemConfig {
+  /// Connection mode definitions that override the built-in modes, keyed
+  /// by mode name: `streaming`, `polling`, or `background`. A definition
+  /// given here replaces the built-in definition for that mode entirely.
+  final Map<String, ModeDefinition> customConnectionModes;
```
kinyoklion: This will change.

packages/common_client/lib/src/config/data_system_config.dart
```diff
@@ -0,0 +1,70 @@
+import '../data_sources/fdv2/mode_definition.dart';
+
+// Maintainer note (not public API): ConnectionModeId is a sealed
+// hierarchy rather than an enum so a custom-mode variant can be added
+// later without changing this surface. The planned extension is a custom
+// variant constructed as `ConnectionModeId.custom('custom-my-mode')`:
```
kinyoklion: We could, of course, append the custom internally.

Then we would add a new version of set connection mode which allowed using ConnectionModeId instead.

packages/common_client/test/data_sources/fdv2/data_system_test.dart
```diff
@@ -0,0 +1,85 @@
...
+  test('an override replaces a built-in mode definition', () {
+    // Override streaming with the polling definition; the streaming
+    // factory should still build a usable data source from it.
+    final factory = makeDataSystem(
+        config: const DataSystemConfig(connectionModes: {
+      ConnectionModeId.streaming: BuiltInModes.polling,
+    })).buildFactories()[const FDv2Streaming()]!;
+
+    final source = factory(_context());
+    expect(source, isA<DataSource>());
```
tanderson-ld: Should this expect the polling source in order to confirm it isn't the streaming one?

packages/common_client/lib/src/config/data_system_config.dart
```diff
@@ -0,0 +1,74 @@
...
+  /// The built-in streaming mode.
+  static const ConnectionModeId streaming = _BuiltInConnectionMode('streaming');
+
+  /// The built-in polling mode.
+  static const ConnectionModeId polling = _BuiltInConnectionMode('polling');
+
+  /// The built-in background mode.
+  static const ConnectionModeId background =
+      _BuiltInConnectionMode('background');
+}
```
tanderson-ld, line 50: Are these going to replace the following in the long term?  I'm ok with this, just trying to wrap my head around all the types in the various resolving layers.

```
        FDv2ConnectionMode,
        FDv2Streaming,
        FDv2Polling,
        FDv2Offline,
        FDv2Background,
```

Perhaps these mode IDs are just for the public configuration layer and they will map in to the FDv2ConnectionMode enum.

tanderson-ld: Following up, those types I listed are currently exported at the Flutter SDK level, and they probably don't need to be if these ConnectionModeId types are the public face.

packages/common_client/lib/src/data_sources/fdv2/data_system.dart
```diff
@@ -0,0 +1,140 @@
...
+        credential: _credential,
+        logger: _logger,
+        httpProperties: _httpProperties,
+        serviceEndpoints: _serviceEndpoints,
+        withReasons: _withReasons,
+        defaultPollingInterval: _defaultPollingInterval,
+        // The common client loads cached flags into the flag store before
+        // the data source starts (FlagManager.loadCached during identify),
+        // so the cache is already applied by the time this chain runs.
+        // Reporting a miss advances the chain without re-applying it.
```
tanderson-ld: This comment seems to "know too much".

packages/common_client/lib/src/data_sources/fdv2/data_system.dart
```diff
@@ -0,0 +1,140 @@
...
+      const FDv2Polling(): _factoryForMode(
+          _resolve(ConnectionModeId.polling, BuiltInModes.polling)),
+      const FDv2Background(): _factoryForMode(
+          _resolve(ConnectionModeId.background, BuiltInModes.background)),
+    };
+  }
+
+  DataSourceFactory _factoryForMode(ModeDefinition modeDefinition) {
+    return (LDContext context) {
+      if (!identical(context, _lastContext)) {
```
tanderson-ld: If you go from context A -> B -> A and there is no decoration via auto env or anonymous context decoration, could this be the same instance?  If this function isn't invoked during the transition (offline mode?), the selector may not get cleared but context B's flags may be loaded from cache.

I'm hand waving a bit and haven't fully investigated.

Android had a context generation bug recently so seeing this makes me wonder if it can happen here.

packages/common_client/lib/src/config/data_system_config.dart
kinyoklion: I can see if I can adjust the FDv2ConnectionMode type to work in a way that supports the needs of configuration. 

Basically it is ideal if you can use them in a map, and that it is easy to do so. I think it maybe can, but the ergonomics may be a bit confusing.

packages/common_client/lib/src/data_sources/fdv2/data_system.dart
kinyoklion: This isn't just used for identify. But I can see if this really needs to even know this.

packages/common_client/lib/src/config/data_system_config.dart
kinyoklion: Removed the FDv2 versions from the export barrel files.

packages/common_client/test/data_sources/fdv2/data_system_test.dart
kinyoklion: They are all the orchestrator type. So it isn't quite that straightforward. It is more an e2e test, but we can make sure it does a poll and make sure it doesn't use SSE.

kinyoklion: The type-to-source mapping (e.g. a polling definition producing a polling source) is covered in `entry_factories_test`. What this test was actually missing is the data system's own job -- confirming the override is *selected* over the built-in; `isA<DataSource>()` passed either way. Updated it to assert `resolvedDefinition(mode)` returns the override for the overridden mode and the built-in for the others.

packages/common_client/lib/src/config/data_system_config.dart
kinyoklion: Removed FDv2ConnectionMode and its subtypes from the Flutter SDK's public exports -- that change lands with the flutter exposure PR rather than here. ConnectionModeId is the public, config-facing id; FDv2ConnectionMode stays the internal resolution type. They are separate on purpose: FDv2ConnectionMode carries value equality, so it cannot key a const config map, whereas the built-in ConnectionModeId values are const-identity and can.

packages/common_client/lib/src/data_sources/fdv2/data_system.dart
kinyoklion: Reworked this so the basis reset no longer keys on the context instance in the factory. It is now driven from the data manager's identify path, keyed on the context canonical key, so it fires on any context change independent of the active mode (including offline) and independent of the factory being invoked at all -- which closes the instance-reuse / offline-transition gap. Added a regression test for the identify-change-identify sequence. See 7f96be5.

packages/common_client/lib/src/data_sources/data_source_manager.dart
```diff
@@ -92,6 +99,21 @@ final class DataSourceManager {
     _activeDataSource = null;
   }
 
+  /// Resolves the pending identify, if any. Idempotent: only the first call
+  /// completes it. Callers decide *when* to call it -- a cached identify on
+  /// the first applied payload, a wait-for-network identify on the
+  /// orchestrator's [InitializedEvent].
```
kinyoklion:
```suggestion
  /// Resolves the pending identify, if any. Idempotent: only the first call
  /// completes it.
```

packages/common_client/lib/src/data_sources/data_source_manager.dart
```diff
@@ -118,6 +140,11 @@ final class DataSourceManager {
 
     switch (_activeConnectionMode) {
       case FDv2Offline():
+        // Report why the SDK is offline. When an offline data source is
+        // configured (the FDv2 data system supplies one) it then loads
+        // cached flags through the pipeline below; its payload does not
+        // drive the status to valid while offline, so this status stands.
+        // FDv1 has no offline factory, so offline stays status-only.
```
kinyoklion:
```suggestion
```

packages/common_client/lib/src/data_sources/data_source_manager.dart
```diff
@@ -146,22 +172,36 @@ final class DataSourceManager {
...
-          // The FDv1 data sources this manager runs never produce FDv2
-          // payload events.
+          var handled = await _dataSourceEventHandler.handlePayload(
+              _activeContext!, event.changeSet,
+              environmentId: event.environmentId);
+          if (handled == MessageStatus.messageHandled) {
+            // Applying any change set from a live source marks it valid --
+            // including a no-change response, which restores valid after an
+            // interruption. While offline the status set in _setupConnection
+            // stands, so cached data does not report a live connection.
```
kinyoklion:
```suggestion
            // Applying any change set from a live source marks it valid --
            // including a no-change response, which restores valid after an
            // interruption.
```

packages/common_client/lib/src/data_sources/data_manager.dart
```diff
@@ -0,0 +1,78 @@
...
+/// cache initializer loads it as the first tier. Identify resolves on the
+/// first delivered payload, or -- when waiting for network results -- only
+/// on fresh data, so a cache load alone does not satisfy a wait-for-network
+/// identify.
+///
+/// Each identify starts data acquisition fresh: any held selector is
+/// discarded via [clearSelector] before connecting, so the new connection
+/// re-fetches a full payload rather than resuming a previous context's
+/// basis. Mode switches keep the selector and reach the data source manager
+/// directly rather than through here, so they are unaffected.
```
kinyoklion:
```suggestion
```

packages/common_client/lib/src/data_sources/data_manager.dart
```diff
@@ -0,0 +1,78 @@
...
+  Future<void> identify(LDContext context,
+      {required bool waitForNetworkResults});
+}
+
+/// FDv1 data manager.
+///
+/// The cache is loaded imperatively at identify time via
+/// [FlagManager.loadCached]. A cache hit resolves identify immediately
+/// unless the caller is waiting for network results; either way the
+/// network connection is started so live data follows.
```
kinyoklion:
```suggestion
```

packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
```diff
@@ -235,30 +261,34 @@ final class FDv2DataSourceOrchestrator implements DataSource {
...
         persist: false,
       ));
     }
+
+    // Initialization completes at exhaustion when there is no synchronizer
+    // to wait on (cache-only), or when an initializer already delivered data
+    // without a selector. Otherwise the first synchronizer completes it.
+    if (_cacheOnlyDataSystem || dataReceived) {
+      _emitInitialized();
+    }
```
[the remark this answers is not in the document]
kinyoklion: Checking this against the other implementations. The wording may be too strong as waiting is a best effort. It doesn't promise you will get a network result.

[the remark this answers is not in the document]
kinyoklion: I lean toward keeping this behavior as it seems like the correct balance of safety.

packages/common_client/lib/src/data_sources/data_source_manager.dart
```diff
@@ -146,22 +177,35 @@ final class DataSourceManager {
...
+            // interruption.
+            if (_activeConnectionMode is! FDv2Offline) {
+              _statusManager.setValid();
+            }
+            // A 'cached' identify resolves on any applied data; a 'fresh'
+            // identify waits for the orchestrator's InitializedEvent
+            // instead.
+            if (_minimumDataAvailability == DataAvailability.cached) {
+              _maybeCompleteIdentify();
+            }
```
tanderson-ld, line 200: I would expect a comparison here like payload.dataAvailable >= _minimumDataAvailability.  Then all paths through which the payload itself completes identify would happen here (and not only cached here, while fresh or exhaustion happens in InitializedEvent).

Having the payload include the data availability could also eliminate the need for the `_cacheOnlyDataSystem` boolean and associated special logic in the orchestrator.

It would also eliminate these situations of back to back signals.

```
          _emitPayload(result);
          _emitInitialized();
```

I think then the InitializedEvent indicates the orchestrator is in steady state, which is a way identify completes as well.

tanderson-ld (approved):

packages/common_client/lib/src/data_sources/data_source_manager.dart
kinyoklion: We aren't directly driving the behavior 1 to 1 with a payload. But what types of signals we have over time. 

Sometimes they are coupled, but other times they are independent. If payloads were always the direct transition, then that would work.

We get a payload, and we are waiting for "fresh" data. 
We handle several more data sources and don't get any better. 
The orchestrator then decides we are initialized. Independent of the payloads.

It would be directly handled in the orchestrator, but we actually have some situations which are outside its control, which is why it is handled at this level.

kinyoklion: Basically the cases for whenever they aren't paired, the cases where they are paired is technically redundant. Otherwise this state is already represented in the selector, which the payload already has.

PR #312, opened 2026-06-24
feat: honor the FDv1 fallback directive on success, error, and goodbye

kinyoklion:
Stacked on #311 (the `SseHttpError` surfacing this depends on).

## What

Honors the FDv1 fallback directive across every way the server can deliver it:

- **Successful connection:** the directive is emitted with the basis change set, so the streamed payload is applied *before* the SDK falls back — previously the basis was dropped the moment the header was seen.
- **Any error response carrying the header** (`SseHttpError`, recoverable or not): the streaming source closes the connection — which stops the client's own retry — and routes to the fallback tier.
- **`goodbye` event with `protocolFallbackTTL`:** treated as an in-band fallback directive, for transports that cannot read response headers.

A single helper parses the directive (presence + TTL) from response headers, used for both the successful and error paths; the goodbye path reads its TTL in-band.

The streaming source maps `SseHttpError` by its `recoverable` flag: recoverable → interrupted (the client retries), unrecoverable → terminal. The FDv1 streaming source ignores recoverable errors so a transient 5xx no longer shuts it down.

When a fallback tier is configured the orchestrator engages it. When none is configured, the SDK stays interrupted and retries FDv2 after the directive's TTL (default 1 hour; a TTL of `0` means remain paused with no retry) rather than halting or reconnecting immediately. Source results carry the fallback TTL.

## Tests

- Orchestrator: apply-then-engage from a directive-bearing change set, and the three no-fallback cases (finite TTL retries, absent TTL defers, zero TTL pauses).
- `streaming_base`: defer-on-success, the TTL header, the goodbye/TTL directive, and the `SseHttpError` paths (recoverable → interrupted/stays open, unrecoverable → terminal/closes, directive → terminal+TTL regardless of recoverability). `protocol_handler` covers `protocolFallbackTTL` parsing.
- v3 contract harness `fdv1-fallback` suite passes end-to-end (816 total, 792 ran, exit 0), with no regression.

The contract-test-service capability + `fdv1Fallback` config wiring that exercises this in the harness lives on the e2e branch and will land with the v3 contract-tests PR.

---

> [!NOTE]
> **Medium Risk**
> Changes core FDv2 connection lifecycle, tier selection, and flag delivery ordering when servers send fallback signals; behavior shifts are broad but covered by new unit/contract tests.
> 
> **Overview**
> Implements end-to-end handling when the server asks the SDK to fall back from FDv2 to FDv1, including **TTL-aware retry** when no FDv1 tier is configured.
> 
> **Directive parsing and propagation:** Adds shared `readFallbackDirective` for `x-ld-fd-fallback` / `x-ld-fd-fallback-ttl`, threads `fdv1FallbackTtl` through `FDv2SourceResult`, and parses in-band `protocolFallbackTTL` on goodbye events.
> 
> **Streaming / polling behavior:** On a **successful** stream open, the directive is **deferred** and emitted with the next change set so the basis payload is applied before fallback (replacing immediate terminal error on connect). **HTTP errors** with the header, and **goodbye** with TTL or a pending header, surface as terminal fallback results so the orchestrator does not recycle past the directive. Polling mirrors goodbye/header TTL stamping. FDv2 streaming maps `SseHttpError` by recoverability; legacy FDv1 streaming **ignores recoverable** SSE HTTP errors so transient 5xx no longer shuts the source down.
> 
> **Orchestrator:** Classifies directives into engage FDv1 tier, defer retry (no tier: wait TTL—default 1h, zero = pause indefinitely), or none; schedules recycle via `_pendingRetryDelay` and interruptible `_delay` on stop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74e7c4bb1f2042178e2bb13f7ad2156ee96a2993. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/common_client/lib/src/data_sources/fdv2/fallback_directive.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/fallback_directive.dart
@@ -1,0 +1,20 @@
+/// The FDv1 fallback directive parsed from a connection's response
+/// headers. Its presence means the server asked the SDK to fall back;
+/// [ttl] is how long to remain on the fallback before retrying FDv2
+/// (null when the server gave no TTL; [Duration.zero] means indefinitely).
+final class FallbackDirective {
+  final Duration? ttl;
+  const FallbackDirective(this.ttl);
+}
+
+/// Reads the FDv1 fallback directive from response headers, or null when
+/// the `x-ld-fd-fallback` header is not `"true"`. The single place that
+/// interprets these headers, shared by the streaming and polling sources.
+FallbackDirective? readFallbackDirective(Map<String, String> headers) {
+  if (headers['x-ld-fd-fallback']?.toLowerCase() != 'true') {
+    return null;
+  }
+  final raw = headers['x-ld-fd-fallback-ttl'];
+  final seconds = raw == null ? null : int.tryParse(raw);
+  return FallbackDirective(seconds == null ? null : Duration(seconds: seconds));
+}
--- a/packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
@@ -34,7 +34,7 @@ enum _SynchronizerOutcome {
   stop,
 }
 
-/// How the orchestrator responds to a server-directed FDv1 fallback.
+/// How the orchestrator responds to server-directed actions.
 enum _DirectiveAction {
   /// No directive present; carry on normally.
   none,
--- a/packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
+++ b/packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
@@ -34,6 +34,20 @@ enum _SynchronizerOutcome {
   stop,
 }
 
+/// How the orchestrator responds to a server-directed FDv1 fallback.
+enum _DirectiveAction {
+  /// No directive present; carry on normally.
+  none,
+
+  /// An FDv1 fallback tier exists and was engaged; advance to it.
+  engageFdv1Fallback,
+
+  /// The directive was received but no FDv1 fallback tier is configured.
+  /// The SDK stays interrupted and retries FDv2 after the directive's TTL
+  /// rather than removing the source.
+  deferRetry,
+}
+
 /// The FDv2 data source orchestrator.
 ///
 /// Runs the initializer chain to bring the SDK to a usable state, then
@@ -63,6 +77,18 @@ final class FDv2DataSourceOrchestrator implements DataSource {
   bool _closed = false;
   bool _emittedPayload = false;
   bool _initialized = false;
+
+  /// Default wait before retrying FDv2 after a fallback directive arrives
+  /// with no FDv1 fallback tier configured and no server-provided TTL.
+  static const Duration _defaultFallbackRetryInterval = Duration(hours: 1);
+
+  /// Overrides the next recycle delay when set (consumed once), used to
+  /// honor a fallback directive's retry TTL.
+  Duration? _pendingRetryDelay;
+
+  /// Completes when [stop] is called so a long interruptible wait (e.g. a
+  /// fallback retry delay) wakes promptly on shutdown.
+  final Completer<void> _stopCompleter = Completer<void>();
 
   /// True when the only sources are cache initializers (no synchronizers).
   /// Such a system must still reach a usable state on a cache miss, so an
@@ -117,6 +143,7 @@ final class FDv2DataSourceOrchestrator implements DataSource {
   void stop() {
     if (_closed) return;
     _closed = true;
+    if (!_stopCompleter.isCompleted) _stopCompleter.complete();
     _sourceManager.close();
     // Wake the active synchronizer run so it can observe the closed
     // state.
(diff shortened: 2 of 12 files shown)


packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
```diff
@@ -34,6 +34,20 @@ enum _SynchronizerOutcome {
   stop,
 }
 
+/// How the orchestrator responds to a server-directed FDv1 fallback.
+enum _DirectiveAction {
```
kinyoklion, line 38: I want the name here to be clear that it is a directive for the FDv1 callback.

packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
```diff
@@ -190,22 +217,51 @@ final class FDv2DataSourceOrchestrator implements DataSource {
...
-  bool _handleFdv1Fallback(FDv2SourceResult result) {
-    if (result.fdv1Fallback &&
-        _sourceManager.hasFdv1FallbackConfigured &&
-        !_sourceManager.isCurrentSynchronizerFdv1Fallback) {
+  /// Classifies a server-directed FDv1 fallback and, for [engaged], makes
+  /// the source manager switch tiers. The directive is ignored
+  /// ([none]) when the FDv1 fallback synchronizer is already the active
+  /// source, so it cannot loop (the fallback source also never re-asserts
+  /// the directive). When a fallback tier is configured it is engaged;
+  /// when none is configured the caller defers an FDv2 retry per the TTL.
```
kinyoklion:
```suggestion
  /// Classifies a server-directed FDv1 fallback and, for [engaged], makes
  /// the source manager switch tiers. The directive is ignored
  /// ([none]) when the FDv1 fallback synchronizer is already the active
  /// source. When there is no FDv1 fallback we defer retrying FDv2 ([deferRetry])].
```

packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
kinyoklion: If we want more directives, then 'engaged' would need to be FDv1 specific instead.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -61,6 +67,14 @@ final class FDv2StreamingBase {
   bool _pingInFlight = false;
   bool _pingPending = false;
 
+  /// Set when a successful connection's response carried the FDv1 fallback
+  /// directive. The directive rides out on the next emitted change set so
```
kinyoklion: Don't use "rides out". Do something more like "The directive will be emitted with the next change set" or something similar.

packages/common_client/lib/src/data_sources/fdv2/streaming_base.dart
```diff
@@ -317,26 +354,47 @@ final class FDv2StreamingBase {
...
-      final fallback = err.headers['x-ld-fd-fallback']?.toLowerCase() == 'true';
-      _logger.warn(
-          'Streaming request failed with status ${err.statusCode}; giving up');
+    if (err is SseHttpError) {
+      final message = 'Streaming request failed with status ${err.statusCode}';
+
+      // A fallback directive on the response takes precedence over the
+      // status: tear down this connection (which stops any retry the
+      // client scheduled) and route to the fallback tier.
+      if (err.headers['x-ld-fd-fallback']?.toLowerCase() == 'true') {
```
kinyoklion: We need to have a centralized helper method for checking this and for getting the the fallback time.

packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
```diff
@@ -34,6 +34,20 @@ enum _SynchronizerOutcome {
   stop,
 }
 
+/// How the orchestrator responds to a server-directed FDv1 fallback.
```
kinyoklion:
```suggestion
/// How the orchestrator responds to server-directed actions.
```

packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
```diff
@@ -34,6 +34,20 @@ enum _SynchronizerOutcome {
   stop,
 }
 
+/// How the orchestrator responds to server-directed actions.
+enum _DirectiveAction {
```
kinyoklion, line 38: We can likely use this for generic retry-after eventually as well.

packages/common_client/lib/src/data_sources/fdv2/orchestrator.dart
```diff
@@ -225,19 +282,26 @@ final class FDv2DataSourceOrchestrator implements DataSource {
...
+              case _DirectiveAction.engageFdv1Fallback:
+                // Data received but the server directed FDv1 fallback;
+                // move on to synchronizers where the fallback tier runs
+                // and its first change set completes initialization.
+                return;
+              case _DirectiveAction.deferRetry:
+                // An initializer is one-shot and cannot retry itself;
+                // let the chain exhaust and leave retry timing to the
+                // synchronizer tier.
+                break;
```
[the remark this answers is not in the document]
kinyoklion: I am skeptical on this one. It does mean, in this specific case, there is an extra request.

packages/common_client/lib/src/data_sources/fdv2/fallback_directive.dart
```diff
@@ -0,0 +1,20 @@
...
+
+/// Reads the FDv1 fallback directive from response headers, or null when
+/// the `x-ld-fd-fallback` header is not `"true"`. The single place that
+/// interprets these headers, shared by the streaming and polling sources.
+FallbackDirective? readFallbackDirective(Map<String, String> headers) {
+  if (headers['x-ld-fd-fallback']?.toLowerCase() != 'true') {
+    return null;
+  }
+  final raw = headers['x-ld-fd-fallback-ttl'];
+  final seconds = raw == null ? null : int.tryParse(raw);
```
joker23, line 18: Question: is there a contract with back end for what the min/max number would be? (eg no negative, or some number that makes us wait for years, etc)

joker23 (approved):

PR #314, opened 2026-06-25
feat: Add experimental support for data-saving mode (FDv2).

kinyoklion:
The final wiring PR for FDv2 in the Flutter SDK: surfaces the data system on the public API so applications can use it. The data system itself lives in `common_client` (already on main via #310/#311/#312); this PR is the Flutter exposure.

## Changes
- **`LDConfig`** gains a `dataSystem` option (passed through to `LDCommonConfig`). Setting it opts the SDK into FDv2. Documented as early access / not semver-stable.
- **`LDClient.setConnectionMode([ConnectionMode?])`** — applies a manual connection-mode override (sticks and suppresses automatic state-detection transitions; `null` clears it and resumes automatic resolution). Early access.
- **Public exports** — adds the data-system configuration types (`DataSystemConfig`, `ConnectionModeId`, `ModeDefinition`, `EndpointConfig`, `InitializerEntry`/`SynchronizerEntry`, `CacheInitializer`/`PollingInitializer`/`StreamingInitializer`/`PollingSynchronizer`/`StreamingSynchronizer`, `Fdv1FallbackConfig`) and removes the internal `FDv2*` connection-mode types from the public surface (they’re used internally by `setConnectionMode`, not part of the API).

## Validation
- `flutter analyze` clean; all 47 flutter package tests pass against current main.
- End-to-end behavior (incl. FDv1 fallback) is exercised by the v3 contract tests — added in the stacked follow-up.

## Release note
The `launchdarkly_common_client` pin is still `1.13.0` (the last released version, which predates FDv2). CI resolves it via the melos path override, so the workspace builds against the FDv2 `common_client` on main. **Before this ships, `common_client` must release with the FDv2 work and this pin must be bumped** (per `RELEASING.md`: common → common_client → flutter, in dependency order).

Two drafts will be stacked on this PR: the v3 contract tests, and the FDv2 example app.

---

> [!NOTE]
> **Medium Risk**
> Changes how connection modes are chosen when FDv2 is enabled and adds experimental runtime overrides over networking/lifecycle behavior; impact is limited by early-access APIs and mostly additive wiring.
> 
> **Overview**
> **Surfaces FDv2 (data-saving mode) on the Flutter public API** by adding optional `dataSystem` on `LDConfig` (passed through to `common_client`) and exporting the related configuration types (`DataSystemConfig`, `ConnectionModeId`, initializer/synchronizer entries, etc.). Internal `FDv2*` connection-mode types are **removed from the public export list**; apps use `ConnectionModeId` instead.
> 
> Adds **`LDClient.setConnectionMode([ConnectionModeId?])`** as an early-access manual override that sticks until cleared with `null`, wired through `ConnectionManager.setMode`. **`ConnectionManagerConfig.initialModeOverride`** seeds that behavior at construction (including from `dataSystem.initialConnectionMode`).
> 
> When **`dataSystem` is configured**, foreground automatic resolution defaults to **streaming** (not FDv1 `dataSourceConfig.initialConnectionMode`), and any configured initial connection mode is applied as the sticky override. FDv1 apps without `dataSystem` keep the previous data-source-driven initial mode.
> 
> Tests cover initial override on startup, lifecycle suppression while overridden, and resuming automatic resolution after clearing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8848521922000ee32c9f687f77f99ec94604a79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/flutter_client_sdk/lib/launchdarkly_flutter_client_sdk.dart
+++ b/packages/flutter_client_sdk/lib/launchdarkly_flutter_client_sdk.dart
@@ -45,11 +45,6 @@ export 'package:launchdarkly_common_client/launchdarkly_common_client.dart'
         PersistenceConfig,
         ApplicationInfo,
         ConnectionMode,
-        FDv2ConnectionMode,
-        FDv2Streaming,
-        FDv2Polling,
-        FDv2Offline,
-        FDv2Background,
         ResolvedConnectionMode,
         ResolvedStreaming,
         ResolvedPolling,
@@ -73,7 +68,19 @@ export 'package:launchdarkly_common_client/launchdarkly_common_client.dart'
         PluginCredentialInfo,
         PluginEnvironmentMetadata,
         PluginMetadata,
-        PollingConfig;
+        PollingConfig,
+        DataSystemConfig,
+        ConnectionModeId,
+        ModeDefinition,
+        EndpointConfig,
+        InitializerEntry,
+        SynchronizerEntry,
+        CacheInitializer,
+        PollingInitializer,
+        StreamingInitializer,
+        PollingSynchronizer,
+        StreamingSynchronizer,
+        Fdv1FallbackConfig;
 
 export 'src/ld_client.dart' show LDClient;
 export 'src/config/ld_config.dart' show LDConfig, ApplicationEvents;
--- a/packages/flutter_client_sdk/lib/src/ld_client.dart
+++ b/packages/flutter_client_sdk/lib/src/ld_client.dart
@@ -78,11 +78,23 @@ interface class LDClient {
         DiagnosticSdkData(name: sdkName, version: sdkVersion),
         hooks: combined);
     final stateDetector = FlutterStateDetector();
+    // Under the FDv2 data system the connection mode is governed by the
+    // data system configuration, not the FDv1 data source options: the
+    // foreground slot defaults to streaming, and an initial connection mode
+    // (if configured) is applied as a sticky override, equivalent to calling
+    // setConnectionMode immediately after construction.
+    final dataSystem = config.dataSystem;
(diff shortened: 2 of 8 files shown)


packages/flutter_client_sdk/lib/launchdarkly_flutter_client_sdk.dart
```diff
@@ -45,11 +45,6 @@ export 'package:launchdarkly_common_client/launchdarkly_common_client.dart'
         PersistenceConfig,
         ApplicationInfo,
         ConnectionMode,
-        FDv2ConnectionMode,
```
kinyoklion, line 48: We technically did export these, but didn't use them. So I am not going to consider it breaking.

packages/flutter_client_sdk/lib/src/ld_client.dart
```diff
@@ -343,6 +343,27 @@ interface class LDClient {
...
+  /// The mode is applied as a manual override: it sticks and suppresses
+  /// automatic state-detection transitions (backgrounding, network
+  /// availability) until cleared. Call with no argument (or null) to clear
+  /// the override and resume automatic mode resolution.
+  ///
+  /// This method is not stable, and not subject to any backwards
+  /// compatibility guarantees or semantic versioning. It is in early
+  /// access. If you want access to this feature please join the EAP.
+  /// https://launchdarkly.com/docs/sdk/features/data-saving-mode
+  void setConnectionMode([ConnectionMode? mode]) {
```
kinyoklion: This isn't using the correct connection modes.

joker23 (approved):

tanderson-ld (approved):

PR #341, opened 2026-09-03
feat: Support usePost for FDv2 polling and streaming requests

kinyoklion:
## Summary

The FDv2 data system carried an unwired `usePost` flag on each mode-definition entry: the requestor and streaming factories honored it, but the built-in modes hardcoded it to `false` and no configuration exposed it. Under FDv2 the SDK always sent the evaluation context in the URL with `GET`, and `useReport` was silently ignored.

- `DataSystemConfig.usePost` (default `false`) switches every FDv2 polling and streaming source to `POST` with the context JSON in the request body, on the `/sdk/poll/eval` and `/sdk/stream/eval` paths. It replaces the per-entry `usePost` fields on `PollingInitializer`, `StreamingInitializer`, `PollingSynchronizer` and `StreamingSynchronizer`, which were not reachable through configuration (the data system is in early access, with no compatibility guarantees). The setting travels through `SourceFactoryContext`, so built-in modes and overrides behave the same.
- On web the browser `EventSource` cannot send a `POST`. When `usePost` is set on such a platform the data system drops the streaming sources from every mode and logs a warning, so a streaming mode polls instead. The platform capability is a new `streamingPostSupported` default next to the existing `streamingReportSupported`.
- `useReport` together with a `DataSystemConfig` now logs a warning that the option is ignored and points at `usePost`.
- The contract test service declares `client-use-post` and maps `clientSide.usePost` onto `DataSystemConfig.usePost`.
- The FDv1 fallback source also keeps the context out of the URL under `usePost`. The FDv1 endpoints do not accept `POST` (production returns 404 for `POST /msdk/evalx/context` and 200 for `REPORT`), so the fallback poller uses `REPORT` with the context JSON body when `usePost` is set, and `GET` otherwise.

The `POST` request variants come from the v3 harness change in launchdarkly/sdk-test-harness#445. With a harness built from that branch the full v3 suite passes against this branch: 844 total, 823 ran, 20 of them `POST` subtests. CI runs the v2 harness, which this change does not affect.

Also verified against production with a mobile key through the contract test service: in polling mode behind a logging proxy the SDK sent `POST /sdk/poll/eval` with the context JSON body and received the environment's flags; in a streaming-only mode (no polling initializer, no cache) with `usePost` the stream alone delivered the same 183 flags, matching the `GET` control runs.

---

> [!NOTE]
> **Overview**
> Adds **`DataSystemConfig.usePost`** (default off) so FDv2 polling and streaming send the evaluation context in a **POST** body instead of a **GET** URL path. The old per-mode-entry `usePost` fields are removed; the setting flows through **`SourceFactoryContext`** to all FDv2 sources.
> 
> On **web**, where `EventSource` cannot POST, enabling `usePost` strips streaming sources from resolved modes, logs a warning, and leaves polling so streaming modes degrade to poll. **`streamingPostSupported`** is added to platform defaults (true on IO, false on JS).
> 
> **FDv1 fallback** honors the same flag via **REPORT** + JSON body (FDv1 does not accept POST). **`useReport`** with an FDv2 data system now logs that it is ignored and points to **`usePost`**.
> 
> The Flutter contract test service advertises **`client-use-post`** and maps harness **`clientSide.usePost`** onto **`DataSystemConfig.usePost`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5671348b6e5287a7ffd910826c4502753840f4c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/common_client/lib/src/data_sources/fdv2/data_system.dart
```diff
@@ -75,19 +91,40 @@ final class FDv2DataSystem {
   };
 
   /// The definition for [mode]: the user's override if one was given for
-  /// it, otherwise the built-in default.
+  /// it, otherwise the built-in default, less any sources the platform
```
joker23, line 94: nit: not understanding this comment change.

joker23 (approved):

packages/common_client/lib/src/data_sources/fdv2/data_system.dart
kinyoklion: There are built-in defaults that get filtered based on what the platform supports.

